PR #2, opened 2012-07-06
Fix for DFL-3310, Syntax error when auto-completing after "in".

chriskr:


chriskr:
Added some tests (there was actually a qunite test already from Rune) and fixed a regression. Will add a link in the manual test suite.

PR #9, opened 2012-07-16
DFL-2826 Return Values

hzr:


src/build-application/build_ecmascript_debugger_6_0.js
```diff
@@ -86,10 +86,17 @@ window.app.builders.EcmascriptDebugger["6.0"] = function(service)
...
-                            ['watches', 'callstack', 'inspection'],
+                            ['watches', 'return-values', 'callstack', 'inspection'],
                             // default expanded flags for the view list
                             [false, true, true]);
 
+    /* Return Values */
+    cls.ReturnValuesView.prototype = ViewBase;
+    new cls.ReturnValuesView('return-values',
+                          ui_strings.M_VIEW_LABEL_RETURN_VALUES, 
+                          'scroll mono');
```
chriskr: Args not lining up properly.

src/ecma-debugger/action_handler.js
```diff
@@ -32,6 +32,13 @@
   }
 };
 
+window.eventHandlers.click["goto-script-line"] = function(event, target)
+{
+  var script_id = Number(event.target.getAttribute("data-script-id"));
+  var script_line = Number(event.target.getAttribute("data-script-line"));
+  views.js_source.show_and_flash_line(script_id, script_line);
+};
```
chriskr, line 40: Why do we need here a new handler? We already have such functionality?

src/ecma-debugger/stop_at.js
```diff
@@ -61,6 +61,7 @@ cls.EcmascriptDebugger["6.0"].StopAt = function()
   var runtime_id = '';
 
   var callstack = [];
+  var return_values = {};
```
chriskr, line 70: This file needs obviously maintenance.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
 
+  this.return_values = function(return_values, search_term)
+  {
+    return (
+      ["ol",
+        return_values.return_value_list.map(function(retval) {
+          return this.return_value(retval,
+                                   return_values.rt_id,
+                                   search_term);
+        }, this),
```
chriskr, line 353: I would let 'return_value' take 'return_values.rt_id' and 'search_term' as first arguments and then use bind instead of a new inline function.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+       "class", "return-values"
+      ]
+    );
+  };
+
+  this.return_value = function(retval, rt_id, search_term)
+  {
+    var search_re = new RegExp(search_term, "ig")
```
chriskr: Missing semicolon. The 'g' flag doesn't matter if you are using it only to test for a match. But i would actually suggest to not use reg exp here. If the search term has special reg exps chars you will get an unexpected behavior (also it's on the TODO list that we will have to support for all these filters and searches ignore case and reg exp).

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+      {
+        value_template.push(
+          ["item",
+            ["value",
+              value,
+             "class", type
+            ]
+          ]
+        );
+      }
```
chriskr: I would try to find a way to not repeat that if block.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+      break;
+
+    case TYPE_OBJECT:
+      var object = retval.value.object;
+      var name = object.className === "Function" && !object.functionName
+               ? ui_strings.S_ANONYMOUS_FUNCTION_NAME
+               : object.functionName;
+      value = window.templates.inspected_js_object(retval.value.model, true, null, search_term);
+      if (value !== "")
+        value_template = [value];
```
chriskr: I all other cases you are pushing the 'value_template' but here you are replacing it. Is that OK? If so it's not really clear why.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+      break;
+    }
+
+    var from_script_id = retval.positionFrom.scriptID;
+    var from_uri = from_script_id && runtimes.getScript(from_script_id)
+                 ? (runtimes.getScript(from_script_id).uri || runtimes.getRuntime(rt_id).uri)
```
chriskr: 'runtimes.getScript' and 'runtimes.getRuntime' can both return 'null'.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+                 : ui_strings.S_UNKNOWN_SCRIPT;
+    var to_script_id = retval.positionTo.scriptID;
+    var to_uri = to_script_id && runtimes.getScript(to_script_id)
+               ? (runtimes.getScript(to_script_id).uri || runtimes.getRuntime(rt_id).uri)
```
chriskr: The same as above.

src/ecma-debugger/templates.js
```diff
@@ -307,6 +339,181 @@
...
+               : ui_strings.S_UNKNOWN_SCRIPT;
+
+    var object = retval.functionFrom;
+    var func_model = new cls.InspectableJSObject(rt_id,
+                                                 object.objectID,
+                                                 object.functionName || ui_strings.S_ANONYMOUS_FUNCTION_NAME,
+                                                 object.className);
+    var func_search_term = value_template.length ? null : search_term;
+    var func = window.templates.inspected_js_object(func_model, true, null, func_search_term);
```
chriskr, line 460: Will you have here not the same issue that any state in the model is not persistent?

OK, after i have read the whole diff i understand better. I think this is actually too expensive to just have that single line without expander. That doesn't require really a model.

src/ecma-debugger/view_return_values.js
```diff
@@ -0,0 +1,131 @@
...
+  this._create_view = function(return_values)
+  {
+    this._container.clearAndRender(templates.return_values(return_values, this._search_term));
+  };
+
+  this._return_values_no_content = function()
+  {
+    return (
+      ["div",
+         "No return values",
```
chriskr: Missing ui string?

src/ecma-debugger/view_return_values.js
```diff
@@ -0,0 +1,131 @@
...
+      this._create_view_bound();
+    }
+  }.bind(this);
+
+  this._init = function(id, name, container_class)
+  {
+    View.prototype.init.call(this, id, name, container_class);
+    this.required_services = ["ecmascript-debugger"];
+    this._container = null;
+    this._models = [];
```
chriskr: It seems this member is not used?

src/ecma-debugger/view_return_values.js
```diff
@@ -0,0 +1,131 @@
...
+    this._search_term = "";
+  };
+
+  this._init(id, name, container_class);
+};
+
+cls.ReturnValuesView.create_ui_widgets = function()
+{
+  new ToolbarConfig
```
chriskr: We now have support for config objects, we should basically deprecate this old form. The new form is more self documenting.

src/ecma-debugger/view_return_values.js
```diff
@@ -0,0 +1,131 @@
...
+        handler: "return-values-text-search",
+        shortcuts: "return-values-text-search",
+        title: ui_strings.S_INPUT_DEFAULT_TEXT_FILTER,
+        label: ui_strings.S_INPUT_DEFAULT_TEXT_FILTER,
+        type: "filter"
+      }
+    ]
+  );
+
+  var text_search = new TextSearch(1);
```
chriskr: I have started to move such thing to the init call of the according view, makes IMO more sense.

src/ui-style/debugger_style.css
```diff
@@ -1607,3 +1607,75 @@ d
...
+}
+
+.return-value .return-value-arrow-to
+{
+  padding-left: 14px;
+}
+
+.return-function-from .folder-key
+{
+  display: none;
```
chriskr, line 1648: This looks a bit like a hack.

src/ui-style/debugger_style.css
```diff
@@ -1607,3 +1607,75 @@ d
...
+.return-values
+{
+  margin: 5px;
+  list-style-type: none;
+  padding: 0;
+}
+
+.return-value
+{
+  display: table;
```
chriskr, line 1619: The consequence of the table design is that prototypes don't look consistent any more with all other views of this model. Is there not a way to address the issue you are solving with this in a different way?

src/ecma-debugger/action_handler.js
hzr, line 40: I can't see that we have such a handler already.

src/ui-style/debugger_style.css
hzr, line 1619: That's what I was trying to fix with this styling. How do they not look like the rest?

src/ecma-debugger/templates.js
hzr, line 353: I tried that, but this this feels like the more logical order for the arguments IMO. I don't want to reorder the arguments based on how I call the function.

src/ecma-debugger/templates.js
hzr: I don't see how that could be much shorter. Do you have something in mind?

hzr:
Added review fixes. cc @chriskr @p01

src/ecma-debugger/action_handler.js
chriskr, line 40: OK, perhaps not exactly in the same way. But breakpoints in Breakpoints and frame in Call Stack hve the same functionality. Also event listener tooltips, but they are not jet on your branch. Script links in the command line too, actually.

src/ui-style/debugger_style.css
chriskr, line 1619: They don't stretch over the whole panel.

src/ecma-debugger/templates.js
chriskr: You could e.g. take TYPE_NUMBER in the block above and make and set 'value' depending on retval.value.type == TYPE_NUMBER.

chriskr:
This looks good now from my point of view. Mathieu should review too.

chriskr:
This now looks good.

PR #16, opened 2012-07-24
Small code cleanup.

hzr:


chriskr:
Looks good.

chriskr:
This should actually go into bugfixes, i think? We never push directly to master, everything must go over cutting-edge first.

hzr:
Yes, that's true.

PR #18, opened 2012-07-24
Small code cleanup.

hzr:


chriskr:
Looks good (ideally each commit to bugfixes has an according BTS number so we can track what we have to test, but i guess in this case it's OK without).

danfooo:
Looks fine

PR #19, opened 2012-07-25
Unprefixed -o-{transform,transition,linear-gradient}.

hzr:
core-integration-364 unprefixes some CSS properties. This updates these properties with the new syntax.

### DIFF
--- a/src/client-en.xml
+++ b/src/client-en.xml
@@ -10,6 +10,8 @@
   background: #e6e6e6 repeat-x -o-linear-gradient(-90deg,
                                                   rgba(255, 255, 255, 0.75) 0,
                                                   rgba(255, 255, 255, 0.25) 100%);
+  background: #e6e6e6 repeat-x linear-gradient(rgba(255, 255, 255, 0.75) 0,
+                                               rgba(255, 255, 255, 0.25) 100%);
   border-radius: 3px;
   line-height: 19px;
   margin: 0 3px;
@@ -28,6 +30,8 @@
   background-image: -o-linear-gradient(-90deg,
                                        rgba(255, 255, 255, 0.45) 0,
                                        rgba(255, 255, 255, 0.25) 100%);
+  background-image: linear-gradient(rgba(255, 255, 255, 0.45) 0,
+                                    rgba(255, 255, 255, 0.25) 100%);
   box-shadow: inset 0 0 0 1px rgba(0, 0, 0, 0.3),
               inset 0 1px 3px rgba(0, 0, 0, 0.35);
 }
--- a/src/console-logger/consolelogger_style.css
+++ b/src/console-logger/consolelogger_style.css
@@ -38,7 +38,7 @@
   padding: 0;
   cursor: pointer;
   -o-transition: -o-transform 0.2s;
-  transition: -o-transform 0.2s;
+  transition: transform 0.2s;
 }
 
 .errors-table > tr.expanded .expander
--- a/src/console-logger/consolelogger_style.css
+++ b/src/console-logger/consolelogger_style.css
@@ -38,11 +38,13 @@
   padding: 0;
   cursor: pointer;
   -o-transition: -o-transform 0.2s;
+  transition: -o-transform 0.2s;
 }
 
 .errors-table > tr.expanded .expander
 {
   -o-transform: rotate(90deg);
+  transform: rotate(90deg);
 }
 
 .errors-table .expand_cell
--- a/src/network/network_templates.js
+++ b/src/network/network_templates.js
@@ -366,6 +366,13 @@ templates.network_graph_sections = function(entry, width, duration, do_tooltip)
                                   "rgba(255, 255, 255, 0.25) 100%), " +
                         "-o-linear-gradient(0deg," +
                           templates.network_graph_sections_style(entry, width, duration) +
+                        ");" +
+                    "background-image: " +
+                        "linear-gradient(0deg," +
+                               "transparent 0%," +
+                               "rgba(255, 255, 255, 0.25) 100%), " +
+                        "linear-gradient(-90deg," +
+                          templates.network_graph_sections_style(entry, width, duration) +
                         ");"
          ];
 };
@@ -458,6 +465,9 @@ templates.network_graph_tooltip = function(entry, mono_lineheight)
             ["div",
               "style", "height: " + HEIGHT + "px; " +
                        "background-image: -o-linear-gradient(270deg," +
+                          templates.network_graph_sections_style(entry, HEIGHT, duration) +
+                        ");" +
+                       "background-image: linear-gradient(180deg," +
                           templates.network_graph_sections_style(entry, HEIGHT, duration) +
                         ");",
               "class", "network-tooltip-graph"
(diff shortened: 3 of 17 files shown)


src/ui-style/tabs.css
```diff
@@ -51,9 +51,11 @@ top-tabs
   height: 30px;
   overflow: hidden;
   background-color: #e9e9e9;
-  background-image: -o-linear-gradient(270deg,
+  background-image: -o-linear-gradient(-90deg,
```
hzr, line 54: This was just for consistency with other values.

src/client-en.xml
```diff
@@ -10,6 +10,8 @@
   background: #e6e6e6 repeat-x -o-linear-gradient(-90deg,
                                                   rgba(255, 255, 255, 0.75) 0,
                                                   rgba(255, 255, 255, 0.25) 100%);
+  background: #e6e6e6 repeat-x linear-gradient(rgba(255, 255, 255, 0.75) 0,
```
hzr, line 13: It now defaults to "to bottom", so I left it out.

chriskr:
Looks good. A single bug fix can be pushed directly to bugfixes, that is a reviewed branch too. A special reviewed branch makes only sense for tasks.

src/console-logger/consolelogger_style.css
```diff
@@ -38,11 +38,13 @@
   padding: 0;
   cursor: pointer;
   -o-transition: -o-transform 0.2s;
+  transition: -o-transform 0.2s;
```
rchl: Did you mean to have transform without -o- here?

BTW. I don't recall us dropping prefixes for transitions/transforms. If we want to prepared for the future, we probably should add all possible combinations because we don't know what will loose prefix first.

src/network/network_templates.js
```diff
@@ -366,6 +366,13 @@ templates.network_graph_sections = function(entry, width, duration, do_tooltip)
                                   "rgba(255, 255, 255, 0.25) 100%), " +
                         "-o-linear-gradient(0deg," +
                           templates.network_graph_sections_style(entry, width, duration) +
+                        ");" +
+                    "background-image: " +
+                        "linear-gradient(0deg," +
+                               "transparent 0%," +
```
rchl, line 372: Specifying 0%, 100% explicitly is not strictly needed. Same in some other places. But I don't mind much if you prefer it that way.

src/resource-manager/resource_style.css
```diff
@@ -122,9 +122,12 @@
 
 .highlighted-line
 {
-  background-image: -o-linear-gradient(90deg, 
+  background-image: -o-linear-gradient(90deg,
                                        #d4e3f4,
                                        #d4e3f4);
+  background-image: linear-gradient(180deg,
```
rchl, line 128: background-color would be sufficient here.

src/ui-style/debugger_style.css
```diff
@@ -1429,6 +1436,12 @@ d
                     -o-linear-gradient(0deg,
                                        #c8c8c8 100%,
                                        #c8c8c8 0);
+  background-image: linear-gradient(90deg,
+                                    #c8c8c8 100%,
+                                    #c8c8c8 0),
+                    linear-gradient(00deg,
```
rchl: "0deg"

src/network/network_templates.js
hzr, line 372: Just copied the old stuff, I'll let the maintainer of the view fix it. :)

src/console-logger/consolelogger_style.css
hzr: Yes, thanks. I'm testing in ci-368 and it only works with the unprefixed versions.

src/ui-scripts/ruler/style.css
```diff
@@ -83,9 +127,31 @@
...
+    -o-linear-gradient(-90deg,
+                       #fff 0,
+                       #fff 1px,
+                       transparent 1px),
+    -o-linear-gradient(-90deg,
+                       #fff 0,
+                       #fff 1px,
+                       transparent 1px);
+  background-image:
+    linear-gradient(180deg,
```
danfooo, line 143: Should it be 0deg here? Prefixed starts from east, so -90deg = north, unprefixed here starts north, so 180deg = south? In other lines further down too.

src/ui-scripts/ruler/style.css
```diff
@@ -162,12 +250,25 @@
...
                        transparent 7px),
-    -o-linear-gradient(-45deg, 
+    -o-linear-gradient(-45deg,
                        transparent 0,
                        transparent 5px,
                        #fff 5px,
                        #fff 7px,
                        transparent 7px);
+  background-image:
+    linear-gradient(45deg,
```
danfooo, line 260: Is 45 correct here? The prefixed one is 45, so it should be 135 if I understand correctly? Further down, -45deg is unchanged in the unprefixed too.

src/ui-style/colorpickerstyle.css
```diff
@@ -97,6 +99,18 @@
                     -o-linear-gradient(45deg,
                                        transparent 75%,
                                        #c8c8c8 75%);
+  background-image: linear-gradient(135deg,
```
danfooo, line 102: Prefixed is -45, so it should be -135 here? Won't check deg further down, maybe I just got it wrong.

src/ui-scripts/ruler/style.css
hzr, line 143: That's what I expected too, but that looked slightly wrong. It might have something to do with the background-position. with 180deg it looks like before.

src/ui-scripts/ruler/style.css
hzr, line 260: This doesn't really matter, we're just painting an "×".

src/ui-style/colorpickerstyle.css
hzr, line 102: It might be painted a bit differently, but the end result is the same.

src/ui-scripts/ruler/style.css
rchl, line 143: Using keywords like 'to bottom', 'to top' is probably less confusing.

-90deg in the old one was in other words 'to top', 180deg in the new one is 'to bottom'.

danfooo, line 143: No, 180deg in the new one is 'from bottom' (and therefor also 'to top'). That's why the unprefixing here is correct.

I'd like to stick with degree values, the fact that the degree value represents the "from" (clockwise, starting north), but the keywords have to be "to xyz" makes it more confusing than helpful.

src/resource-manager/resource_style.css
p01, line 128: Agreed.
This class name sound ( and should be ) very generic. However I see it only used in the resource_detail view, to highlight a specific line when viewing a text/\* resource from another service ( e.g. Documents )

danfooo, line 128: Wonder if there's a reason for the gradient? Also prefixed "90deg" is unprefixed "0deg".

danfooo, line 128: Checked, there is background-position and background-size set inline, that's why this needs to be a background-image and therefor a gradient. Maybe the name is a bit too generic, but this is correct. I have nox fixed the degree value, but that's just for consistency, it of course doesn't make a difference.

PR #22, opened 2012-07-26
Hide Return Values section from old versions.

hzr:


danfooo:
Looks good

chriskr:
Looks good.

PR #23, opened 2012-07-27
Task dfl 3295 network details overhaul

danfooo:


src/build-application/build_resource_manager_1_0.js
```diff
@@ -2,6 +2,7 @@
 
 window.app.builders.ResourceManager["1.0"] = function(service)
 {
+  window.network_logger = new cls.NetworkLoggerService();
```
hzr: Why create a global?

src/ecma-debugger/helpers.js
```diff
@@ -377,6 +377,14 @@ window.cls.Helpers = function()
...
    };
   })();
+
+  this.eq = function(prop, val)
+  {
+    return function(obj)
+    {
+      return obj[prop] === val;
+    };
+  }
```
hzr: Missing semicolon.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
 
 window.templates || (window.templates = {});
+window.templates.network || (window.templates.network = {});
 
 (function(templates) {
 
-templates.network_detail_row = function(wrap)
+var HTTP_BOUNDARY_CLASS = "http-token-type-boundary";
+
+templates._wrap_col_or_row = function(wrap)
```
hzr: `wrap` sounds like a weird name here, no? (And yea, I'm nit-picking.)

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    "style", "left:" + left_val + "px"]
+  );
 };
 
-templates.network_log_detail = function(entry)
+templates._details_content = function(entry, do_raw)
 {
-  var responsecode = entry.responses.last && entry.responses.last.responsecode;
+  // Bind a template function for raw / not-raw, on demand.
+  var template_func_name = "_requests_responses_" + (do_raw ? "raw" : "not_raw" + "_bound");
```
hzr: Do you mean `((do_raw ? "raw" : "not_raw") + "_bound")`?

I don't get it though, why create these functions in this way instead of up-front?

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+  if (!this[template_func_name])
+    this[template_func_name] = this.requests_responses.bind(this, do_raw);
+
+  var requests_responses = entry.requests_responses.map(this[template_func_name]);
+  if (do_raw)
+  {
```
hzr: Can skip braces.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-            ["th", ui_strings.M_NETWORK_REQUEST_DETAIL_STATUS + ":"],
-            ["td",
-              entry.touched_network && responsecode ? String(responsecode) : ui_strings.S_RESOURCE_ALL_NOT_APPLICABLE
-            ],
-           "data-spec", "http#" + entry.responsecode
+    ["table",
+      ["tbody",
+        this._wrap_col_or_row(
+          [
+            "h1",
```
hzr: Usually on the same line as the bracket, line 78 too.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+      entry.touched_network ? [] : this.did_not_touch_network(entry),
+      requests_responses
     ]
   );
 };
 
-templates.network_response = function(response)
+templates.did_not_touch_network = function(entry)
+{
+  var local = [
```
hzr: (Sorry, misunderstood at first, deleted those comments.)

I would suggest `local_url_types` or something for clarity.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+        ["p", local.contains(entry.urltype) ? ui_strings.S_NETWORK_NOT_REQUESTED
+                                            : ui_strings.S_NETWORK_SERVED_FROM_CACHE])
+    ]);
+};
+
+templates.requests_responses = function(do_raw, request_response, index, requests_responses)
 {
+  var template_func = this._request;
+  if (request_response.is_response)
+    template_func = this._response;
```
hzr: Might as well do

``` javascript
var template_func = request_response.is_response ? this._response : this._request;
```

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-templates.request_details = function(req)
+templates._response = function(response, is_last_response, do_raw)
 {
-  var ret = [];
-  if (!req || req.urltype === cls.ResourceManager["1.2"].UrlLoad.URLType.DATA)
-    return ret;
+  return [
+    this._response_headers(response, do_raw),
+    this._response_body(response, do_raw, is_last_response)
+  ]
```
hzr: Missing semicolon (also on line 139).

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    {
+      if (state.data_spec_firstline_tokens.contains(state.firstline_tokens))
+      {
+        // Add data-spec attributes on certain firstline tokens, tracked in state
+        attrs.extend(["data-spec", "http#" + (token[STR]).trim()]);
+      }
+      state.firstline_tokens++;
+    }
+    return ["span", token[STR]].concat(attrs);
   }
```
hzr, line 163: Missing semicolon.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
   {
-    if (req.requestbody && req.requestbody.partList && req.requestbody.partList.length)
-      ret.push(templates.network_detail_row(["h2", ui_strings.S_NETWORK_MULTIPART_REQUEST_TITLE]));
+    var TYPE = 0;
+    var STR = 1;
+    var attrs = ["class", "header-token-type-" + cls.HTTPHeaderTokenizer.classnames[token[TYPE]]];
+
+    if (token[TYPE] === cls.HTTPHeaderTokenizer.types.NAME)
+    {
+      attrs.extend(["data-spec", "http#" + (token[STR]).trim()]);
```
hzr, line 160: Nitpicky, but the parens around `token[STR]` are unnecessary.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    }
     else
-      ret.push(templates.network_detail_row(["h2", ui_strings.S_NETWORK_REQUEST_DETAIL_REQUEST_TITLE]));
+    if (token[TYPE] === cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART)
```
hzr: Move up after the `else`.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-  {
-    tbody.push(templates.network_detail_row(ui_strings.S_NETWORK_REQUEST_NO_HEADERS_LABEL));
-  }
-  else
+templates._token_receiver = function(tokens, token_type, token)
+{
+  tokens.push([token_type, token]);
+};
+
+templates.TokenStateholder = function(data_spec_firstline_tokens)
```
hzr: Not sure if this belongs in the templates.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
       {
-        firstline = [
-          ["span", parts[0] + " ", "data-spec", "http#" + parts[0]],
-          ["span", parts[1] + " "],
-          ["span", parts[2] + " "]
+        req.header_tokens = [];
+        var tokenizer = new cls.HTTPHeaderTokenizer();
+        tokenizer.tokenize(req.request_headers_raw, this._token_receiver.bind(this, req.header_tokens));
+      }
+      if (req.header_tokens.length)
```
hzr, line 185: Newline before.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+  if (method_str)
+    method_str = " – " + method_str;
+
+  if (req.requestbody && req.requestbody.partList && req.requestbody.partList.length)
+    ret.push(["h2", ui_strings.S_NETWORK_MULTIPART_REQUEST_TITLE + method_str]);
+  else
+    ret.push(["h2", ui_strings.S_NETWORK_REQUEST_DETAIL_REQUEST_TITLE + method_str]);
+
+  if (!req.request_headers)
+  {
```
hzr: Might drop the braces in this if-else, since the one above doesn't have any.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-      ["span", parts[1], "data-spec", "http#" + parts[1]],
-      ["span", resp.firstline.slice(parts[0].length + parts[1].length + 1)]
-    ];
+    if (!resp.header_tokens)
+    {
+      resp.header_tokens = [];
+      var tokenizer = new cls.HTTPHeaderTokenizer();
+      tokenizer.tokenize(resp.response_headers_raw, this._token_receiver.bind(this, resp.header_tokens));
+    }
+    if (resp.header_tokens.length)
```
hzr, line 232: Newline before.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-{
-  var lis = headers.map(function(header) {
-      return ["tr", ["th", header.name + ":"], ["td", header.value], "data-spec", "http#" + header.name];
-  });
+  var ret = [];
+  var responsecode = resp.responsecode || "";
+  if (responsecode)
+  {
+    if (responsecode in cls.ResourceUtil.http_status_codes)
+      responsecode = responsecode + " " + cls.ResourceUtil.http_status_codes[responsecode];
```
hzr: A bit more compact:

``` javascript
var status_code = cls.ResourceUtil.http_status_codes[responsecode];
if (status_code)
  responsecode = responsecode + " " + status_code;
```

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-templates.network_body_seperator = function()
+templates.headers_list = function(headers, do_raw)
 {
-  return ["pre", " ", "class", "mono"];
+  var map_func;
+  if (do_raw) // This is just for headers of multipart-parts.
+  {
+    map_func = function(header)
+    {
+      return this._wrap_pre([["span", header.name + ":", "data-spec", "http#" + (header.name).trim()], ["span", " " + header.value]]);
```
hzr: Long line. Skip the parens around `header.name`.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
 
-  var cont = [];
-  // when partList.length is 0, the request was one that did not send data
-  if (req.requestbody.partList.length)
+  var ret = [this._wrap_pre("\n")];
+  if (req.requestbody.partList.length) // Multipart
   {
+    var use_raw_boundary = false;
+    if (do_raw && req.boundary)
+      use_raw_boundary = true;
```
hzr: `var use_raw_boundary = Boolean(do_raw && req.boundary);`. (Guessing `req.boundary` is not a boolean.)

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
     }
   }
   else if (req.requestbody.mimeType.startswith("application/x-www-form-urlencoded"))
   {
-    var parts = req.requestbody.content.stringData.split("&");
-    var tab = [
-                ["tr",
+    if (do_raw)
+    {
+      ret.push(this._wrap_pre(req.requestbody.content.stringData, "network-body"));
```
hzr: Does `req.requestbody.content` always exist? Also on line 342.

src/network/network_service.js
```diff
@@ -480,6 +545,20 @@ cls.RequestContextPrototype = function()
       return "uid-" + count++;
     }
   })();
+
+  this.discard_incomplete_warning = function(window_id)
+  {
+    for (var i = 0, window_context; window_context = this.window_contexts[i]; i++)
```
hzr, line 551: Use braces for the `for`.

src/network/network_service.js
```diff
@@ -498,26 +577,20 @@ cls.NetworkLoggerEntry = function(id, resource_id, document_id, context_starttim
...
   this.events = [];
   this.event_sequence = [];
-  this.is_fetching_body = false;
+  this.requests_responses = [];
+  this.current_request = null;
+  this.current_response = null;
+  this.current_responsecode = null;
+  this.called_get_body = false;
+  this.set_is_finished_on_responsefinished = false;
+  this._requestID = 0;
```
hzr: This might be for consistency, but we should use `_request_id` for our own code.

src/network/network_service.js
```diff
@@ -901,29 +972,113 @@ cls.NetworkLoggerEntryPrototype = function()
...
 
   this.__defineGetter__("duration", function()
   {
     return (this.events.length && this.endtime - this.starttime) || 0;
   });
+  this.__defineSetter__("duration", function(){});
+
+  this.__defineGetter__("waiting_time", function()
+  {
+    var waiting_time = this.event_sequence.filter(helpers.eq("classname", CLASSNAME_WAITING))
```
hzr, line 1003: We should access globals with `window.`. Maybe just add `var helpers = window.helpers;`.

src/network/network_service.js
```diff
@@ -901,29 +972,113 @@ cls.NetworkLoggerEntryPrototype = function()
...
 cls.NetworkLoggerEntry.prototype = new cls.NetworkLoggerEntryPrototype();
 
+
+cls.NetworkLoggerRequest = function(entry)
+{
+  this.method = null;
+  this.request_headers = null;
+  this.request_headers_raw = null;
+  this.request_type = null;
+  this.requestbody = null;
```
hzr: All others are named `request_*`, why is this not?

src/network/network_service.js
```diff
@@ -901,29 +972,113 @@ cls.NetworkLoggerEntryPrototype = function()
...
+  this.boundary = "";
+  this.was_responded_to = false;
+  // Set from template code, when first needed:
+  this.header_tokens = null;
+  // Belongs here, unused though:
+  this.requestID = entry.requestID;
```
hzr: Same as before, should be `request_id`.

src/network/network_service.js
```diff
@@ -901,29 +972,113 @@ cls.NetworkLoggerEntryPrototype = function()
...
+  this._update_event_request = function(event)
+  {
+    this.method = event.method;
+  };
+
+  this._update_event_requestheader = function(event)
+  {
+    this.request_headers = event.headerList;
+
+    for (var n=0, header; header = this.request_headers[n]; n++)
```
hzr: `n = 0` with spaces.

src/network/network_templates.js
```diff
@@ -96,8 +97,22 @@ templates.network_request_crafter_main = function(url, loading, request, respons
...
+  {
+    return [];
+  }
+
+  // Only add the title if there's more than one context in total
+  var title = "";
+  if (all_contexts.length > 1)
+  {
+    var win = window.window_manager_data.get_window(context.id);
+    title = win && (" - " + win.title);
```
hzr: We usually have an en dash for these kind of things: "–".

src/network/network_view.js
```diff
@@ -228,20 +228,20 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler)
...
         renderer: function(entry) {
-          return (entry.responsecode && String(entry.responsecode)) || "";
+          return (entry.current_responsecode && String(entry.current_responsecode)) || "";
         },
         title_getter: function(entry, renderer) {
-          if (cls.ResourceUtil.http_status_codes[entry.responsecode])
-            return String(cls.ResourceUtil.http_status_codes[entry.responsecode]);
+          if (cls.ResourceUtil.http_status_codes[entry.current_responsecode])
+            return String(cls.ResourceUtil.http_status_codes[entry.current_responsecode]);
           return renderer(entry);
```
hzr: I'd prefer

``` javascript
var responsecode = cls.ResourceUtil.http_status_codes[entry.current_responsecode];
return responsecode ? String(responsecode) : renderer(entry);
```

src/scripts/spec_links.js
```diff
@@ -2012,6 +2012,7 @@ SpecLinks.specs = {
     "Content-MD5": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.15",
     "Content-Range": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.16",
     "Content-Type": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.17",
+    "Content-Disposition": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.5.1",
```
hzr: I'd insert this in name order.

src/syntaxhighlight/http-header/tokenizer.js
```diff
@@ -0,0 +1,123 @@
...
+  var LINEBREAK_CHARS = {
+    "\r": 1, // CR
+    "\n": 1  // LF
+  };
+
+  this.tokenize = function(input_buffer, ontoken)
+  {
+    this._state_handler = this._state_handlers.FIRST_LINE_PART;
+    this._buffer = input_buffer;
+    this._emitToken = ontoken;
```
hzr, line 29: `_emit_token`

src/syntaxhighlight/http-header/tokenizer.js
```diff
@@ -0,0 +1,123 @@
...
+
+      var c = this._buffer.charAt(this._current_pos++);
+      this._token_type = cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART;
+      if (c in WHITESPACE_CHARS)
+      {
+        this._emitToken(this._token_type ,this._token_buffer);
+        this._token_buffer = "";
+      }
+      else
+      if (c in LINEBREAK_CHARS)
```
hzr: Move up after `else`.

src/syntaxhighlight/http-header/tokenizer.js
```diff
@@ -0,0 +1,123 @@
...
+    FIRST_LINE_PART: function()
+    {
+      if (this._is_EOF())
+        return false;
+
+      var c = this._buffer.charAt(this._current_pos++);
+      this._token_type = cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART;
+      if (c in WHITESPACE_CHARS)
+      {
+        this._emitToken(this._token_type ,this._token_buffer);
```
hzr: Space on the wrong side of the comma (also in other places).

src/syntaxhighlight/http-header/tokenizer.js
```diff
@@ -0,0 +1,123 @@
...
+  NAME            : 2,
+  VALUE           : 3,
+  PUNCTUATOR      : 4
+};
+
+cls.HTTPHeaderTokenizer.classnames = {
+  1 : "first_line_part",
+  2 : "name"           ,
+  3 : "value"          ,
+  4 : "punctuator"
```
hzr: Unusual formatting for these two objects above. Don't line up the values in the first one, and the commas in the second one, and no space before the colon.

In order to always keep them in sync, it would also be nicer to do:

``` javascript
cls.HTTPHeaderTokenizer.classnames = {};
cls.HTTPHeaderTokenizer.classnames[FIRST_LINE_PART] = "first_line_part";
...
```

hzr:
Review done, mainly small stuff.

src/ecma-debugger/runtimes.js
```diff
@@ -1332,14 +1332,15 @@ cls.EcmascriptDebugger["6.0"].Runtimes = function(service_version)
     return  __scripts[scriptId] && __scripts[scriptId].runtime_id || null;
   }
 
-  this.reloadWindow = function()
+  this.reloadWindow = function(passed_window)
```
chriskr: Just window_id as parameter would be a bit more precise.

src/ecma-debugger/runtimes.js
```diff
@@ -1332,14 +1332,15 @@ cls.EcmascriptDebugger["6.0"].Runtimes = function(service_version)
...
   {
-    if (__selected_window)
+    var sel_window = passed_window || __selected_window;
```
chriskr: __selected_window gets the default value for the suggested window_id parameter, so you don't need here a new var.

src/ecma-debugger/views.js
```diff
@@ -477,6 +477,7 @@ cls.MainView .create_ui_widgets = function()
 
   eventHandlers.click['reload-window'] = function(event, target)
   {
-    runtimes.reloadWindow();
+    var window_id = target.get_attr("parent-node-chain", "data-reload-window-id");
+    runtimes.reloadWindow(window_id);
```
chriskr, line 481: window_id should be a of type number (it will work with a string too though).

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+templates._wrap_col_or_row = function(wrap)
 {
+  // Wraps either ["elem", "text"] in a column
+  // or [["elem", "text"], ["elem", "text"]] in a row.
+  if (Array.isArray(wrap[0]) && wrap[1])
+  {
+    return ["tr", wrap[0], wrap[1]];
+  }
   return ["tr", ["td", wrap, "colspan", "2"]];
 };
```
chriskr, line 20: This is a bit a strange function. Seems also to require that the first item of the nested list case is actually "td" or "th"?

After having read the templates i think it could make sense to just split this function into two and use them appropriately. Also as we are in templates I'm not sure if you need the wrap prefix (the same for the wrap_pre).

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+  var is_last_of_type = true;
+  for (var i = index + 1, req_res; req_res = requests_responses[i]; i++)
+  {
+    if (request_response.is_response == req_res.is_response)
+    {
+      is_last_of_type = false;
+      break;
+    }
+  }
+  return template_func.call(this, request_response, is_last_of_type, do_raw);
```
chriskr: AFAICS the code would be more readable if you would do the request_response.is_response check here and call the according template directly.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    this._response_headers(response, do_raw),
+    this._response_body(response, do_raw, is_last_response)
+  ]
+};
 
-  if (req.touched_network)
+templates._make_header_token_templ_func = function(state)
+{
+  return function(token)
   {
```
chriskr: The pattern here is partial invocation. I would try to do that always with bind, e.g. have the returned function signature as state, token and then bind it with state where you need it.

src/network/network_details_templates.js
chriskr: I cannot see the whole line, but this is basically just an other template? Why do you have to define it inline?

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    {
+      return this._wrap_pre([["span", header.name + ":", "data-spec", "http#" + (header.name).trim()], ["span", " " + header.value]]);
+    }.bind(this);
+  }
+  else
+  {
+    map_func = function(header)
+    {
+      return [["th", header.name + ":", "data-spec", "http#" + (header.name).trim()], ["td", header.value]];
+    };
```
chriskr: The same here. Also a map function doesn't need to be bound, just pass the according context in the map call.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
-    cont.push(tab);
+                ]); // It's necesary to just push the outer array, because each entry will be wrapped in a row.
+
+      ret.extend(parts.map(function(e) {
+                    e = e.replace(/\+/g, "%20").split("=");
+                    return [
+                        ["td", decodeURIComponent(e[0])],
+                        ["td", decodeURIComponent(e[1])]
+                    ];
+                  }));
```
chriskr: This too, it's basically just an other template, i don't see why this is defined inline.

src/network/network_details_templates.js
```diff
@@ -1,289 +1,445 @@
...
+    }
   }
   else // not multipart or form.
   {
-    var tpl = [];
     if (req.requestbody.content)
     {
       var type = cls.ResourceUtil.mime_to_type(req.requestbody.mimeType);
-      switch (type)
+      if (["markup", "script", "css", "text"].contains(type))
```
chriskr: Depending how often this gets called it would probably make sense to create for this list a constant.

src/network/network_service.js
```diff
@@ -395,20 +426,29 @@ cls.RequestContextPrototype = function()
...
-      Checks if the event's requestID is different from the one in last_entry.
+      Checks if the event's requestID is different from the one in current_entry.
       That should never be the case, since the "urlload" event initiates
-      a new entry and that doesn't have a requestID. Note that last_entry is
+      a new entry and that doesn't have a requestID. Note that current_entry is
       the last entry we saw with the event's resourceID.
     */
     return event.requestID &&
-           (last_entry.requestID !== event.requestID);
+           (current_entry._requestID !== event.requestID);
```
chriskr: Looks wrong. You should never access a private member on something else than this.

src/network/network_service.js
chriskr: It seems this is the only private member? Is there any reason for that?

src/network/network_service.js
```diff
@@ -728,111 +794,116 @@ cls.NetworkLoggerEntryPrototype = function()
...
-      this.responses.push(this._current_response);
+      if (this.current_request)
+      {
+        this.current_request.was_responded_to = true;
+      }
+      this.current_response = new cls.NetworkLoggerResponse(this);
+      this.requests_responses.push(this.current_response);
     }
-    this._current_response._update_event_responseheader(event);
+    this.current_response._update_event_responseheader(event);
```
chriskr: As above, looks wrong to call a private member on something else than this.

src/network/network_service.js
```diff
@@ -728,111 +794,116 @@ cls.NetworkLoggerEntryPrototype = function()
...
   };
 
   this._update_event_responsefinished = function(event)
   {
-    if (this._current_response)
-      this._current_response._update_event_responsefinished(event);
+    if (this.current_response)
+      this.current_response._update_event_responsefinished(event);
```
chriskr: Here too.

src/network/network_service.js
```diff
@@ -842,8 +913,8 @@ cls.NetworkLoggerEntryPrototype = function()
     else
       this.type = cls.ResourceUtil.mime_to_type(this.mime);
 
-    if (this._current_response)
-      this._current_response._update_mime_and_type(this.mime, this.type);
+    if (this.current_response)
+      this.current_response._update_mime_and_type(this.mime, this.type);
```
chriskr: Private member access.

src/network/network_service.js
chriskr, line 1003: Also seeing how this is used eq should definitely cache the returned functions.

src/network/network_templates.js
chriskr: Also title can here obviously be null or undefined. That will get stringified later, it seems?

src/syntaxhighlight/http-header/tokenizer.js
```diff
@@ -0,0 +1,123 @@
...
+        return false;
+
+      var c = this._buffer.charAt(this._current_pos++);
+      this._token_type = cls.HTTPHeaderTokenizer.types.VALUE;
+      // LINEBREAK_CHARS only mean switching to header when the following char is not whitespace.
+      if (c in LINEBREAK_CHARS && !(this._buffer.charAt(this._current_pos) in WHITESPACE_CHARS))
+      {
+        this._emitToken(this._token_type ,this._token_buffer);
+        this._token_buffer = "";
+        this._state_handler = this._state_handlers.NAME;
```
chriskr, line 88: I would add here a comment that LINEBREAK_CHARS are added by purpose to NAME tokens.

src/build-application/build_resource_manager_1_0.js
danfooo: The idea is that the Resource UI and the Network Crafter can make use of that instance.

hzr: But there's probably a better way to achieve that than to make a global.

src/network/network_details_templates.js
danfooo: ah, `(do_raw ? "raw" : "not_raw") + "_bound"` is what I mean, thanks. I do it this way so it's "lazy" – but it of course doesn't make a difference, I can change it if you want? Also I have renamed `not_raw` to `parsed`.

hzr: I'd change it. :) There's no real benefit, only makes it harder to read.

src/network/network_details_templates.js
danfooo: No, maybe not. Moving it to cls.HTTPHeaderTokenizer.TokenStateholder for now, I think that's better. It's an asset that only makes sense to use with the tokenizer.

src/network/network_details_templates.js
danfooo: I don't. Fixed.

src/network/network_details_templates.js
danfooo: No, it's not. Nice. It would work the same way without Boolean, but `use_raw_boundary` would be a string then, I guess it's cleaner.

src/network/network_details_templates.js
danfooo: Yes, this is within _request_body, I'm pretty sure we'll always have at least an empty string there.

src/network/network_service.js
danfooo: No, that's wrong. There are a few more that can be private now. Have changed this now, `_current_request`, `_current_response`, `_set_is_finished_on_responsefinished`and `_called_get_body` can be private too. Also some properties were missing there and were just set later, I hope I found them all now.

src/network/network_service.js
danfooo, line 1003: I'm not sure how to do that. I could store the function by id, but that would bind the value too. Storing it by id + val is hard because val possibly can't be stringified. Those cases where it can be stringified could be cached I suppose, but it would need a try/catch to find out.

src/network/network_service.js
danfooo: I don't see a reason either, think I just didn't change it. Fixing now.

src/network/network_templates.js
danfooo: What do you mean, "That will get stringified later"? No, I think you're right, when it's "" I'd rather show nothing than the dash. I didn't realize it could be null too, but that's caught too then. Changed line to title = win && win.title && (" – " + win.title);

src/network/network_view.js
danfooo: Much better, yes.

src/syntaxhighlight/http-header/tokenizer.js
danfooo, line 29: I did this to be consistent with other tokenizers.

src/syntaxhighlight/http-header/tokenizer.js
danfooo: I stole the style from cls.CSSTokenizer.types, but yes, no point in copying that. Fixing. Putting them in there one by one I find ugly though, repeating just to align is strange to me. This looks like generated code. We do it in other places too though, so if it's policy, I'll fix.

src/ecma-debugger/runtimes.js
danfooo: Okay, I was a bit careful with this code. So I can `__selected_window = window_id || __selected_window;`and work with __selected_window afterwards?

src/syntaxhighlight/http-header/tokenizer.js
hzr: I wasn't referring to alignment, I meant that if you use the constant as the property name, the name and the constant value will always be in sync, even if you re-order the values.

src/network/network_details_templates.js
danfooo, line 20: Yes, it only makes sense if "td" or "th" is the first item of the list. The reason why it's not two function is that I can do e.g. 

``` javascript
   if (!req.request_headers)
    ret.push(ui_strings.S_NETWORK_REQUEST_NO_HEADERS_LABEL);
  else
    ret.extend(this.headers_list(req.request_headers));
```

and later use `map` with that function. headers_list can have either returned a `pre` with content (raw) or a `[[th], [td]]` row (parsed).

danfooo, line 20: Changed the name to _col_or_row

src/network/network_details_templates.js
danfooo: I think I did it to repeat less, but I agree it's much shorter and more readable when called directly.

src/network/network_details_templates.js
danfooo: True, yes.

src/network/network_details_templates.js
danfooo: "The same here", I assume you mean making these into two actual functions and just decide which one to pass to map. Changing to that. I think I sometimes try to avoid making that many named functions, but I understand it's more efficient than making the function every time.

Passing this to the map call makes sense, done that too.

src/network/network_details_templates.js
danfooo: I think I just left that as is, but fixing now.

src/network/network_details_templates.js
danfooo: True, it's also used one more time. Added.

src/network/network_service.js
danfooo: It's still within the network_service. Okay, but I'll change that.

danfooo: It's a bit weird to have `request_id` just next to things like `resource_id` and `document_id`, because `resource_id` is almost like a constant, as soon as it's set it will never change. `request_id` is just for mapping the right events to the entry, and can change.

But I guess `resource_id` and `document_id` aren't constants either, since their final value is not defined in the constructor, right?

src/network/network_service.js
danfooo: Okay, it's more clear here. I have changed all functions on `Response` to are public. I had also changed `current_response` to private, there are a few cases where I want to access it from the service, but I think I'll make getters and only use them in this case.

src/build-application/build_resource_manager_1_0.js
danfooo: We discussed it over lunch. Passing it now.

src/network/network_details_templates.js
danfooo: Fixing

src/syntaxhighlight/http-header/tokenizer.js
danfooo: Aha, sorry, read that a bit quick. I see what you mean.. changing it now, maybe it's a bit hard to read, but since I need `FIRST_LINE_PART` from `cls.HTTPHeaderTokenizer.classnames`, I'll pass that and `cls.HTTPHeaderTokenizer.classnames` into an anonymous function for now.

src/network/network_service.js
chriskr, line 1003: This is quite hot code, i guess?  If you don't know how to cache then store 'helpers.eq("classname", CLASSNAME_WAITING)' in a scope variable, it returns always the same function.

src/network/network_templates.js
chriskr: Later in the code you concatenate title with an other string, that will cast title to a string, so if win.title is null it will add the string "null"? I think it's safer to set here title to the empty string in that case instead.

src/ecma-debugger/runtimes.js
chriskr: I meant the other way around, if (!window_id) window_id = __selected_window; and then work only with windw_id in that function. __selected_window should not be set here, the method name doesn't suggest that.

src/network/network_service.js
chriskr: A private member starting with underscore should always only be accessed on the this, everything else is basically wrong. But if request_id is only something which you have define as internal helper, then it would perhaps make sense to expose an according method to check what you need to check here? I don't know the details of your design well enough to say if that would make sense or not though.

src/ecma-debugger/runtimes.js
danfooo: Ah, makes sense now. Fixing.

src/network/network_service.js
danfooo: It's accessed and checked a lot, so i think it wouldn't be worth it. And what I'd want to express is not that this is actually private. I'd more like to make `resource_id`into `RESOURCE_ID`, but as I said it's not set from the start. I'll leave it like that, I should maybe actually write some small bit of documentation how it works.

hzr:
Looks fine.

PR #25, opened 2012-07-30
Danfooo bugfixes Bug fix package 8

danfooo:


src/cookie-manager/cookie_manager_views.js
```diff
@@ -430,6 +430,15 @@ cls.CookieManager.CookieManagerViewBase = function()
       // "domain" is val of [input] (with add_cookie service present), or runtimes .hostname
       var domain       = domain_input && domain_input.value.trim() || runtime && this.data._rts[runtime].hostname;
 
+      // Fix expires value, work around CORE-47780: .value property of <input type=datetime-local>
+      // element has two digits representing milliseconds, instead of three.
+      if (expires.split(".").length > 1)
```
hzr: Might split only once (instead of this and line 437).

hzr:
Except for the one comment, it looks good.

danfooo:
Okay, also one if statement does it.

hzr:
Looks fine!

src/cookie-manager/cookie_manager_views.js
```diff
@@ -430,6 +430,15 @@ cls.CookieManager.CookieManagerViewBase = function()
       // "domain" is val of [input] (with add_cookie service present), or runtimes .hostname
       var domain       = domain_input && domain_input.value.trim() || runtime && this.data._rts[runtime].hostname;
 
+      // Fix expires value, work around CORE-47780: .value property of <input type=datetime-local>
+      // element has two digits representing milliseconds, instead of three.
+      var expires_parts = expires.split(".");
+      if (expires_parts[1] && expires_parts[1].length < 3)
```
chriskr: You would get that check a bit cheaper with a reg exp.

danfooo: Is that just a remark or does it block? :)

danfooo: For this a case I don't see much benefit in a regular expression. I find it quicker to read as is.

hzr:
``` javascript
var expires_parts = expires.split(".");
if (expires_parts[1] && expires_parts[1].length < 3)
```

versus e.g.

``` javascript
if (!/\..{3}/.test(expires))
```

I think I prefer the shorter one, but do what you want.

danfooo:
Alright, I was thinking too complicated. It has to be a positive test though, since the .value can skip milliseconds completely. That's when there was no value attribute to begin with and the user sets a date-and-time.

hzr:
Looks fine.

danfooo:
@chriskr @hzr Added one more fix.

chriskr:
Looks good AFAICT.

danfooo:
@hzr yes, when expires is 0 it means the cookie is valid for the session, I forgot that. That cases actually need another fix it seems.

danfooo:
One more fix. Had introduced a regression by dealing with timezone-offset only when reading the value from DOM, not when writing it. Fixed in the latest commit, I hope this is done now.

hzr:
Good!

danfooo:
One more try. @hzr @chriskr

danfooo:
One more bugfix (65b6570) and review fixes. @hzr @chriskr

hzr:
All fine!

src/scripts/dom.js
```diff
@@ -777,6 +777,39 @@ Date.prototype.toLocaleISOString = function()
  return new Date(this.getTime() - this.getTimezoneOffset() * 1000 * 60).toISOString().replace('Z','');
 };
 
+Date.fromLocaleISOString = function(localeISOString)
```
hzr: I guess this should be on the prototype btw.

danfooo:
See last commit, I moved the workaround for unparsable date-strings out of toLocaleISOString, I think this also makes it clearer where what problem is solved.

I'm against caching anything though, if toLocaleISOString needs to be used at all (or if we suddenly get a UTC string) won't change during a session, but the check is cheap and I'd rather look at the value and decide every time. This is all only used when cookies are edited, not worth storing anything per session.

hzr:
Looks good now.

danfooo:
@hzr @chriskr ^^ one more attempt..

hzr:
Looks fine!

chriskr:
Yes, looks good.

p01:
Goody

PR #28, opened 2012-08-02
Task dfl 3042 ev listener

chriskr:


src/ui-scripts/cells.js
```diff
@@ -292,9 +292,9 @@
     {
       ["tabs", "tabbar"].forEach(function(prop)
       {
-        if (rough_cell[prop] && typeof rough_cell[prop] == "function")
+        if (rough_cell["get_" + prop] && typeof rough_cell["get_" + prop] == "function")
```
hzr: This could simply be `if (typeof rough_cell["get_" + prop] == "function")`.

Maybe also

``` javascript
var getter = rough_cell["get_" + prop];
if (typeof getter == "function")
  rough_cell[prop] = getter(services);
```

hzr:
One comment, otherwise fine!

danfooo:
Looks fine.

chriskr:
@hzr updated with your suggestion

hzr:
A-ok++.

PR #31, opened 2012-08-07
DFL-3371 Make 'configure message filters' more accessible.

rchl:
The dialog for configuring message filters is moved to the side
of the view rather than being a popup dialog.

Also store last value entered in the message sending view so that
it's not cleared on switching views.


src/debug/debug.js
```diff
@@ -104,11 +208,17 @@ cls.debug.Debug = function(id, name, container_class)
     }
   };
 
+  this._filter_command = function(msg){return /^handle/.test(msg);}
```
hzr: Some whitespace missing, should be `function(msg) { return /^handle/.test(msg); }`.

Same on following lines.

src/debug/debug.js
```diff
@@ -188,6 +298,67 @@ cls.debug.Debug = function(id, name, container_class)
     this.update();
   };
 
+  this.show_configuration = function(container, service)
+  {
+    container.render(this._template_service_config(service));
+  }
+
+  this.set_unfold = function(service, bool)
```
hzr: `bool` could be more descriptive (I know this follows old code though).

hzr:
Otherwise fine, taking into account that there is a lot of old code that was just moved.

src/debug/debug.js
rchl: Should I change it in the "interface" section at the top also?

hzr: Yea.

hzr:
Looks fine!

src/debug/debug.js
```diff
@@ -104,11 +208,17 @@ cls.debug.Debug = function(id, name, container_class)
...
+  this._filter_event = function(msg, index) { return /^on/.test(msg); }
+
+  this._get_message_class = function(msg) { return msg.replace(/^handle/, "").replace(/^on/, "On"); }
+
   /* implementation */
 
   this.createView = function(container)
   {
-    this._textarea = container.clearAndRender(this._main_template());
+    this._textarea = container.clearAndRender(this._main_template()).childNodes[1];
```
chriskr: querySelector("textarea") would be a bit more solid towards changes.

src/lib/test_framework.js
```diff
@@ -351,6 +352,8 @@ window.cls.TestFramework = function()
                 {'pre': pres[1]}
               )
           }
+          var txt = message_container.querySelector('textarea');
+          txt.oninput = (function() { this._last_txtarea_value = txt.value; }).bind(this);
```
chriskr, line 356: Normally we use our event delegation design, but i guess an exception for this view is OK.

chriskr:
Yes, looks fine.

hzr:
Looks fine!

chriskr:
Yes, looks fine.

PR #34, opened 2012-08-07
Chrisk bugfixes

chriskr:


src/lib/storageprototype.js
```diff
@@ -0,0 +1,19 @@
...
+  var raw_value = this.getItem(key);
+  var value = null;
+  try
+  {
+    value = JSON.parse(raw_value);
+  }
+  catch(e)
+  {
+    value = default_value;
+  };
```
hzr: Extra semicolon.

src/lib/storageprototype.js
```diff
@@ -0,0 +1,19 @@
...
+  return value === null || value === undefined ? default_value : value;
```
hzr: `return value == null ? default_value : value;` would suffice, but this is fine.

src/ui-scripts/settings.js
```diff
@@ -32,12 +31,10 @@ var SettingsBase = function()
    */
   this.get = function(key)
   {
-    var val = "";
-    return (
-    typeof this.map[key] !== 'undefined' ?
-    this.map[key] :
-    (this.map[key] = ((val = window.localStorage.getItem(key)) ? JSON.parse(val) : null)));
-  }
+    if (typeof this.map[key] == "undefined")
```
hzr: No need for `typeof` here. And if that disappears, maybe we can check against `null` (to catch that too if needed):

``` javascript
if (this.map[key] == null)
```

hzr:
Looks fine otherwise.

danfooo:
Looks fine, I didn't look deep enough to know what checks are really needed where, but there are comments on that already.

src/lib/storageprototype.js
p01: Agreed.

p01:
All good beside the null/undefined checks that can be combined.

chriskr:
Simplified boolean check.

hzr:
Looks good!

PR #39, opened 2012-08-13
DFL-3447 Browser JS has no longer a label in the JS dropdown in Scripts

chriskr:
Regression caused by the fixes in the URI class, relative URIs no longer have a filename.


danfooo:
Looks fine.

p01:
Agreed

PR #40, opened 2012-08-13
Rechecked the converted unprefixed-gradient degree numbers, corrected some

danfooo:


danfooo:
The according bug report: https://bugs.opera.com/browse/DFL-3450

hzr:
I'm fine with this.

chriskr:
Well, i guess it's OK, but i would prefer if would focus on real issues. We have enough of them.

PR #41, opened 2012-08-13
DFL-3453 JS drop-down does not show the name/uri of the selected script ...

chriskr:
...when you select a User JS or Injected Script or Browser JS


danfooo:
Looks fine

hzr:
Good!

PR #42, opened 2012-08-14
Chrisk bugfixes

chriskr:


hzr:
Fine otherwise.

chriskr:
Committed review fixes.

hzr:
Looking nice!

src/cookie-manager/cookie_manager_data.js
```diff
@@ -131,29 +145,35 @@ cls.CookieManager.CookieDataBase = function()
...
-        // runtime was not active and is to be removed from this._rts
-        delete this._rts[rt_id];
+        // rt is a string, rt_id is a number which can now be compared with what's in msg.runtimes_with_dom
+        var rt_id = this._rts[rt].rt_id;
+        if (msg.runtimes_with_dom.indexOf(rt_id) === -1)
+        {
+          // runtime was not active and is to be removed from this._rts
+          delete this._rts[rt_id];
+          this.cookie_list = this.cookie_list.filter(function(cookie)
+          {
```
danfooo, line 169: Inline function, should now have `{` on the same line. Only old code though.

src/cookie-manager/cookie_manager_data.js
```diff
@@ -252,96 +272,94 @@ cls.CookieManager.CookieDataBase = function()
     }
   };
 
-  this._handle_cookies = function(status, message, rt_id, active_tab_counter)
+  this._handle_cookies = function(status, message, rt_id)
   {
-    if (this._active_tab_count === active_tab_counter)
+    var rt = this._rts[rt_id];
+    if (!rt)
+      throw "missing runtime in _handle_cookies";
```
danfooo, line 279: Would that be possible? To me it looks like we wouldn't have made it that far, did you add that because the last active-tab message is stored and may be outdated? Still would be a value here AFAICS.

src/ecma-debugger/objectinspection.6.0/inspectablejsobject.js
```diff
@@ -320,22 +319,21 @@ cls.EcmascriptDebugger["6.0"].InspectableJSObject.prototype = new function()
   this._get_all_ids = function get_all_ids(tree, ret)
   {
     ret || (ret = []);
-    if (tree)
+    if (tree && tree.get("object_id"))
     {
-      ret.push(tree.object_id);
-      for (var index in tree.protos)
+      ret.push(tree.get("object_id"));
```
danfooo, line 324: That `tree.get` is done in the check too, maybe worth storing?

src/cookie-manager/cookie_manager_data.js
chriskr, line 279: Well, it shouldn't, that's why it throws here :) . About what is possible and what not, I'm just careful with asynchronous code.

src/ecma-debugger/objectinspection.6.0/inspectablejsobject.js
chriskr, line 324: I don't think so. It's a method, but it's more or less the same as doing `foo.bar`, it has almost the same performance.

src/ecma-debugger/objectinspection.6.0/inspectablejsobject.js
```diff
@@ -125,26 +125,28 @@ cls.EcmascriptDebugger["6.0"].InspectableJSObject.prototype = new function()
...
-      if (!tree.protos[index])
-        tree.protos[index] = {};
+      var key = path[i][PATH_KEY];
+      var obj_id = path[i][PATH_OBJ_ID];
+      var index = path[i][PATH_PROTO_INDEX];
+      if (i < (path.length - 1) && !(tree.get_chain(["protos", index, key])))
+        throw "not valid path in InspectableJSObject._get_subtree";
+
+      if (!tree.get("protos"))
+        tree.set("protos",  new Dict());
```
danfooo: Extra space before `new Dict()`

danfooo:
Few random comments, looks fine otherwise.

src/cookie-manager/cookie_manager_data.js
chriskr, line 169: I'm not aware of that rule.

hzr, line 169: It's been in the style guide for a long time, but I know we've discussed it since. Brace on the next line keeps it consistent, but it's not lining up with the actual function which is a bit weird. I don't feel strongly, but maybe we should make a definitive decision.

chriskr:
@hzr @danfooo there was one more commit, have you reviewed that too?

danfooo:
Reviewed it now, few comments on the commit.

hzr:
Fine otherwise.

chriskr:
Pushed some review fixes.

hzr:
Looks fine.

danfooo:
Looks fine.

PR #43, opened 2012-08-16
One fix, one icon.

danfooo:


hzr:
Looks good.

p01:
Goody.
I'm still a little confused myself about .set_query_selector() method :p

danfooo:
The elements in the container matching that querySelector are searched, I guess based on their textContent. Gets really tricky when you have matches that contain other matches, the search goes crazy then. Most of the time I think it's good to select only one level deep in such a selector.. but that didn't make sense here either.

PR #45, opened 2012-08-17
Bugfix package 9

hzr:
Plus some small fixes.


danfooo:
Looks fine except for 1 comment. On 032eec9, I have probably seen that bug, would be good to update experimental with this soon.

src/lib/elementprototype.js
```diff
@@ -24,7 +24,7 @@ if (!Element.prototype.insertAdjacentHTML)
   }
 }
 
-if (typeof document.createElement('div').classList == 'undefined')
+if (!document.createElement('div').classList)
 {
```
chriskr: I think we should remove this altogether. It's an  incomplete implementation of `classList`, ci 167 has `classList`, and further down in this file we check for `document.createElement('div').classList` again. AFAICS that following code would actually fail because of this incomplete implementation here.

chriskr:
Other than that it looks good.

hzr:
One more commit.

chriskr:
Looks good, especially for a Saturday night. :)

hzr:
@chriskr @danfooo More.

danfooo:
Other than that it's good!

hzr:
Ok, one more.

danfooo:
Looks good. I don't think we should add it, it's probably better to get used to the actual api. And we didn't need it so far.

hzr:
I actually think we should add support for several classes. It makes sense and the API may be changed: https://www.w3.org/Bugs/Public/show_bug.cgi?id=13999

chriskr:
Yes looks good. Would not make an API change now.

PR #47, opened 2012-08-20
First part of make request task + Interface for Resource UI

danfooo:


src/network/network_request_crafting_view.js
```diff
@@ -76,8 +81,10 @@ cls.RequestCraftingView = function(id, name, container_class, html, default_hand
     var requestline = lines.shift();
     var reqparts = requestline.match(/(\w*?) (.*) (.*)/);
 
-    if (!reqparts || reqparts.length != 4) {
-        return null; // fixme: tell what's wrong
+    if (!reqparts || reqparts.length != 4)
```
hzr: Old code, but I don't know why 4 is significant here. A comment would help.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+      var is_main_context = type === this.CONTEXT_TYPE_MAIN;
+      ctx = this._contexts[type] = new cls.RequestContext(this, is_main_context);
+      this.post("context-added", {"context_type": type});
     }
+    return ctx;
+  };
+
+  this.remove_request_context = function(type)
+  {
+    type = (type || this.CONTEXT_TYPE_MAIN);
```
hzr: Would drop the parens.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+    var crafter = this._contexts[this.CONTEXT_TYPE_CRAFTER];
+    if (crafter)
     {
-      var window_context = new cls.NetworkLoggerService.WindowContext(data.windowID);
-      this._current_context.window_contexts.push(window_context);
-      if (!data.parentDocumentID)
+      if (crafter.is_waiting_for_create_request)
+      {
+        // Store in a queue. Before we know what resourceID create_request
+        // will return, messages can't be assorciated with the right context.
```
hzr: "associated"

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+        this._message_queue.push([listener, msg]);
+        return true;
+      }
+      else
       {
-        window_context.saw_main_document = true;
+        // Play back the message queue.
+        while (this._message_queue.length)
+          this._playback_message_bound(this._message_queue.shift());
+
```
hzr: Extra line.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+    var ctx = this.get_request_context(this.CONTEXT_TYPE_LOGGER, true);
+
+    // Without a parentDocumentID, this event means "unload" for the old content of this windowID.
+    if (!data.parentDocumentID)
+      ctx.remove_window_context(data.windowID);
+
+    var window_context = ctx.get_window_context(data.windowID, true);
+    if (!data.parentDocumentID)
+      window_context.saw_main_document = true;
+
```
hzr, line 106: Extra line.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+  };
+  this._on_abouttoloaddocument_bound = this._on_abouttoloaddocument.bind(this, this._on_abouttoloaddocument_bound);
```
hzr: Is passing in the function you're binding as the second argument intentional here? Unless I'm missing some context, it will always be `undefined` at that point.
Edit: ah, I guess you want to pass `this._on_abouttoloaddocument`.

danfooo: Correct, thanks.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+      return;
 
     var data = new cls.ResourceManager["1.0"].UrlRedirect(msg);
+    var ctx = this._get_matching_context(data.fromResourceID);
+    if (!ctx)
+      return;
+
+    // Allocate the resource_id we redirect to to the same context.
+    if (data.fromResourceID in ctx.allocated_res_ids)
+      ctx.allocated_res_ids[data.toResourceID] = ctx.allocated_res_ids[data.fromResourceID];
```
hzr, line 140: This could be

``` javascript
var res_id = ctx.allocated_res_ids[data.fromResourceID];
if (res_id)
  ctx.allocated_res_ids[data.toResourceID] = res_id;
```

Both are fine, but this is slightly easier to read IMO.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
-    this._current_context.update("response", data);
-  }.bind(this);
+    var ctx = this._get_matching_context(data.resourceID);
+    if (!ctx)
+      return;
 
-  this._on_request_bound = function(msg)
+    ctx.update("response", data);
+  }
+  this._on_response_bound = this._on_response_bound.bind(this, this._on_response_bound);
```
hzr: Hmm, does `this._on_response_bound` already exist at this point? Or should it be `this._on_response_bound = this._on_response.bind(this, this._on_response);`?

src/network/network_request_crafting_view.js
danfooo: Yes, .match returns the whole match as [0], will actually slice it off and work only with the parts.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+        remove_from_allocated_after_update = true;
+      }
+    }
 
-  this._on_urlunload_bound = function(msg)
+    ctx.update("responsefinished", data);
+
+    if (remove_from_allocated_after_update)
+      delete ctx.allocated_res_ids[data.resourceID];
+
```
hzr, line 280: Extra line.

src/network/network_service.js
```diff
@@ -221,95 +369,85 @@ cls.NetworkLoggerService = function()
...
+  }
 
-  this.unpause = function()
+  this.get_resources = function(resource_ids)
   {
-    if (this._current_context)
-      this._current_context.unpause();
+    var filter_bound = this._filter_entries.bind(this, resource_ids);
+    var entries = this._context.get_entries().filter(filter_bound);
+    return entries.map(function(entry){return new cls.ResourceInfo(entry)});
```
hzr: Some missing whitespace and semicolon.

src/network/network_service.js
```diff
@@ -221,95 +369,85 @@ cls.NetworkLoggerService = function()
...
-  this.entry_ids = [];
-}
+cls.NetworkLoggerService.WindowContext.prototype = new cls.NetworkLoggerService.WindowContextPrototype();
 
-cls.RequestContext = function()
+cls.RequestContext = function(service, is_main_context)
 {
+  this.FILTER_ALLOW_ALL = {
+    type_list: [],
+    "is_blacklist": true
```
hzr: Any reason why this is quoted?

src/network/network_service.js
```diff
@@ -493,16 +624,51 @@ cls.RequestContextPrototype = function()
...
         // Store the id in the list of entries in the window_context
-        var window_context = this.get_window_context(event.windowID);
+        var window_context = (event.windowID && this.get_window_context(event.windowID, true));
         window_context.entry_ids.push(id);
       }
       logger_entry.request_id = event.requestID;
+
+      // Add a mapped crafter_request_id when applicable
+      if (res_id in this.allocated_res_ids)
+        logger_entry.crafter_request_id = this.allocated_res_ids[res_id];
```
hzr, line 655: Here too you can do:

``` javascript
var crafter_request_id = this.allocated_res_ids[res_id];
if (crafter_request_id)
  logger_entry.crafter_request_id = crafter_request_id;
```

(Assuming the id can't be 0 here.)

src/network/network_service.js
danfooo, line 140: The values in `allocated_res_ids` are not resource ids, but (my own) entry ids. The keys are resource ids. So it's more a dict of resource ids for the context, within the service the value may as well always just be `true`.  The value I put in now is only relevant to the crafter (in the next step, not now actually).

I think I like it better as it is now, IMO it expresses that dict character better, and the code doesn't assume what is actually kept in there, it just copies it over so the redirect will have the right context associated.

src/network/network_service.js
```diff
@@ -493,16 +624,51 @@ cls.RequestContextPrototype = function()
...
+    }
+  };
+
+  this.post_on_context_or_service = function(name, body)
+  {
+    // Find out where to post the update message.
+    // Messages of main_contexts are posted on the service, not the context.
+    var posting_object = this;
+    if (this._is_main_context)
+      posting_object = this._service;
```
hzr: `var posting_object = this._is_main_context ? this._service : this;` is short and clear.

src/network/network_service.js
```diff
@@ -546,19 +712,39 @@ cls.RequestContextPrototype = function()
...
+    var request = [
+      windowid,
+      url,
+      requestdata.method,
+      requestdata.headers,
+      null, // payload
+      3, // header policy. 2 == overwrite, 3 == replace
+      2, // reload policy. 2 == no cache, always reload from network
+      null, // request content mode
+      [1, 1] // response content mode 1 == string, 1 == decodee
```
hzr: "decode"

src/network/network_service.js
```diff
@@ -546,19 +712,39 @@ cls.RequestContextPrototype = function()
...
+    ];
+    this.is_waiting_for_create_request = true;
+    var id = this._get_uid();
+    var tag = window.tagManager.set_callback(null, this._handle_create_request.bind(this), [id]);
```
hzr: Write `window.tag_manager` instead.

src/build-application/build_resource_manager_1_0.js
```diff
@@ -12,7 +12,12 @@ window.app.builders.ResourceManager["1.0"] = function(service)
                          null,
                          "network-logger",
                          logger_service);
-  new cls.RequestCraftingView("request_crafter", ui_strings.M_VIEW_LABEL_REQUEST_CRAFTER, "scroll", "", "");
+  new cls.RequestCraftingView("request_crafter",
+                              ui_strings.M_VIEW_LABEL_REQUEST_CRAFTER,
+                              "scroll",
+                              "",
+                              "",
+                              logger_service);
```
chriskr: The name `logger_service` is a bit confusing here, i think. A service is basically a dynamically created scope interface of one of the existing services, like e.g. ResourceManager, exposed as `window.services["resource-manager"]`. Is it really that?

src/network/network_service.js
danfooo: Correct, of course I mean to always bind the unbound function. A bit surprised I didn't see this failing, I thought I saw a case where we have to play back a  "response" message because sendRequest took quite long to reply. But it would have thrown, so I guess that wasn't the case.

src/network/network_view.js
```diff
@@ -650,49 +653,74 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
+      this.update();
+    }
+  }.bind(this);
 
-  eh.click["toggle-raw-cooked-response"] = this._on_clicked_toggle_response_bound;
-  eh.click["toggle-raw-cooked-request"] = this._on_clicked_toggle_request_bound;
-  eh.click["clear-log-network-view"] = this._on_clear_log_bound;
+  this._init = function(id, name, container_class, html, default_handler)
+  {
+    var eh = window.eventHandlers;
```
hzr: `window.event_handlers` instead.

src/network/network_request_crafting_view.js
```diff
@@ -4,13 +4,13 @@
  * @constructor
  * @extends ViewBase
  */
-cls.RequestCraftingView = function(id, name, container_class, html, default_handler) {
+cls.RequestCraftingView = function(id, name, container_class, html, default_handler, service)
+{
+  this._service = service;
```
chriskr: The same here: is this really a service as we usually use services?

src/network/network_view.js
```diff
@@ -650,49 +653,74 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
+    eh.click["clear-log-network-view"] = this._on_clear_log_bound;
 
-  var contextmenu = ContextMenu.get_instance();
-  contextmenu.register("network-logger-context", [
-    {
-      label: ui_strings.S_CLEAR_NETWORK_LOG,
-      handler: this._on_clear_log_bound
-    }
-  ]);
+    messages.addListener("single-select-changed", this._on_single_select_changed_bound);
```
hzr, line 695: `window.messages`.

src/network/network_view.js
```diff
@@ -650,49 +653,74 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
+    this._service.addListener("context-removed", this._on_context_removed_bound);
+    this._service.addListener("resource-update", this.update.bind(this));
+
+    ActionHandlerInterface.apply(this);
+    this._handlers = {
+      "select-next-entry": _make_selection_func.bind(this, "nextElementSibling"),
+      "select-previous-entry": _make_selection_func.bind(this, "previousElementSibling"),
+      "close-details": this._on_clicked_close_bound
+    };
+    this.id = id;
```
hzr: Maybe move this to the beginning of `init()`, looks random here.

src/network/network_request_crafting_view.js
```diff
@@ -47,10 +46,16 @@ cls.RequestCraftingView = function(id, name, container_class, html, default_hand
 
   this._render_main_view = function(container)
   {
+    var ctx = this._service.get_request_context(this._service.CONTEXT_TYPE_CRAFTER);
+    var entries = [];
+    if (ctx)
+      entries = ctx.get_entries_filtered();
```
chriskr: As an API this is IMO a bit unclear: filtered by what?

src/network/network_service.js
danfooo: I doubt it. Fixing.

src/network/network_request_crafting_view.js
```diff
@@ -132,56 +139,24 @@ cls.RequestCraftingView = function(id, name, container_class, html, default_hand
...
-    }
-    else
-    {
-      this._prev_response = ui_strings.S_INFO_REQUEST_FAILED;
-      this.update();
-    }
-  }.bind(this);
+    var CONTEXT_TYPE_CRAFTER = this._service.CONTEXT_TYPE_CRAFTER;
+    // todo: the old contexts will probably be kept for comparing previous requests.
+    this._service.remove_request_context(CONTEXT_TYPE_CRAFTER);
```
chriskr: The method `this._service.get_request_context(this._service.CONTEXT_TYPE_CRAFTER)` should IMO return a context of type CRAFTER. The according method should IMO take the returned context as argument, e.g. `var crafter_ctx = this._service.get_request_context(this._service.CONTEXT_TYPE_CRAFTER)` and to remove it `this._service.remove_request_context(crafter_ctx)`.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
 
 cls.NetworkLoggerService = function()
 {
-  this._current_context = null;
+  this.CONTEXT_TYPE_LOGGER = this.CONTEXT_TYPE_MAIN = "network-logger";
+  this.CONTEXT_TYPE_CRAFTER = "request-crafter";
```
chriskr: I think these should be static members of the class (and just a number would actually be enough)..

src/network/network_service.js
danfooo, line 655: No, it can't be. Actually crafter_request_id is missing on the Entry when it's initialized, fixing that too. Also adding a check if crafter_request_id was already set, that only makes sense to do once.

src/build-application/build_resource_manager_1_0.js
danfooo: No, it is not. It's the data model, but then the namespace where it lives, `cls.NetworkLoggerService` is wrong too. I guess it could just be NetworkLogger? I think we mostly have a name containing "view" for the views.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+        // will return, messages can't be assorciated with the right context.
+        this._message_queue.push([listener, msg]);
+        return true;
+      }
+      else
       {
-        window_context.saw_main_document = true;
+        // Play back the message queue.
+        while (this._message_queue.length)
+          this._playback_message_bound(this._message_queue.shift());
```
chriskr: AFAICS the queue gets only emptied if there is a crafter. Should it not get emptied without a crafter too?

src/network/network_view.js
danfooo: Adding `var messages = window.messages;` a bit further up.

src/network/network_service.js
hzr, line 140: `s/res_id/id` then. The code is semantically equivalent (well, almost. This is also assuming `id` can't be `0`). But do what you want. :)

src/network/network_request_crafting_view.js
danfooo: True, it doesn't make sense for the crafter either. `get_entries` is the one to use here, filtered is about type and xhr filters, they don't apply here. I think the API is okay for the actual logger UI, makes more sense in that context.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+  this._playback_message_bound = function(queued)
+  {
+    var LISTENER =  0;
+    var MSG = 1;
+    queued[LISTENER].call(this, queued[LISTENER], queued[MSG], true);
   }.bind(this);
 
-  this._on_urlload_bound = function(msg)
+  this._on_abouttoloaddocument = function(listener, msg, is_playing_back)
   {
```
chriskr, line 89: Not sure if i fully understand, but it looks like you could get that easier with some like `this._on_abouttoloaddocument = function listener (msg, is_playing_back)` then `listener` will only be visible in that scope and refer to the function itself.

src/network/network_request_crafting_view.js
danfooo: ATM I keep only one context by type, this way a context itself doesn't need to know what type it is. But it probably can probably make sense to keep a series of crafter context as stated in the comment, so for that it has to work like you suggest. It was a simplification but I might as well fix it now.

src/network/network_service.js
danfooo: Yes, true.

src/network/network_service.js
danfooo, line 89: True, it's much better this way, using on_abouttoloaddocument as the name though as that also shows up in a callstack. So that's one more advantage.

src/network/network_service.js
danfooo: But when making a context from the outside you can now refer to it, as in  `this._logger.get_request_context(this._logger.CONTEXT_TYPE_CRAFTER)`. I can make them numbers though, I had to get used to this, but we do it in all kinds of places.

src/network/network_request_crafting_view.js
danfooo: It's a bit difficult and actually unnecessary to keep more than one context of a type. So I think remove_request_context should still only take `type`. I can change it so there are separate add and remove methods per type, so the type is in the name. They will still need to take a `force` flag though, as the createView methods use them to check if there's a context or not, without requiring one.

src/network/network_service.js
```diff
@@ -2,141 +2,285 @@
...
+  {
+    if (!is_playing_back && this._queue_message(listener, msg))
+      return;
 
     var data = new cls.ResourceManager["1.0"].UrlRedirect(msg);
+    var ctx = this._get_matching_context(data.fromResourceID);
+    if (!ctx)
+      return;
+
+    // Allocate the resource_id we redirect to to the same context.
```
chriskr: to to

danfooo: That is what I want to say. But maybe it reads easier with a comma between the two.

src/network/network_service.js
```diff
@@ -546,19 +712,39 @@ cls.RequestContextPrototype = function()
...
+  {
+    var windowid = window.window_manager_data.get_debug_context();
+    var request = [
+      windowid,
+      url,
+      requestdata.method,
+      requestdata.headers,
+      null, // payload
+      3, // header policy. 2 == overwrite, 3 == replace
+      2, // reload policy. 2 == no cache, always reload from network
```
chriskr: This is IMO easier to read with constants, like `HEADER_POLICY_REPLACE` and `HEADER_POLICY_NO_CACHE`, then you don't need the comments.

src/network/network_service.js
chriskr: Same here, constants makes it easier to read.

src/network/network_service.js
danfooo: I just moved that from the crafter, but yes, can fix that.

src/network/network_service.js
chriskr: No need to bind here, the first argument is the calling target.

src/network/network_service.js
```diff
@@ -546,19 +712,39 @@ cls.RequestContextPrototype = function()
...
+    window.services["resource-manager"].requestCreateRequest(tag, request);
+    return id;
   };
 
-  this.get_window_context = function(window_id)
+  this._handle_create_request = function(status, msg, id)
   {
-    return this.window_contexts.filter(helpers.eq("id", window_id))[0];
+    this.is_waiting_for_create_request = false;
+    if (status == 0)
```
chriskr: I think we should start to use `status == SUCCESS`, makes it easier to read.

src/network/network_view.js
```diff
@@ -650,49 +653,74 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
+
+    eh.click["close-incomplete-warning"] = this._on_close_incomplete_warning_bound;
+
+    this._service.addListener("context-added", this._on_context_established_bound);
+    this._service.addListener("context-removed", this._on_context_removed_bound);
+    this._service.addListener("resource-update", this.update.bind(this));
+
+    ActionHandlerInterface.apply(this);
+    this._handlers = {
+      "select-next-entry": _make_selection_func.bind(this, "nextElementSibling"),
```
chriskr: `_make_selection_func.bind(this,` looks strange: either have a `_make_something_func` or use `bind`, but i cannot imagine that you need both.

src/network/network_view.js
```diff
@@ -650,49 +653,74 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
+    eh.click["type-filter-network-view"] = this._on_change_type_filter_bound;
+    eh.click["profiler-mode-switch"] = this._on_toggle_network_profiler_bound;
+
+    eh.click["close-incomplete-warning"] = this._on_close_incomplete_warning_bound;
+
+    this._service.addListener("context-added", this._on_context_established_bound);
+    this._service.addListener("context-removed", this._on_context_removed_bound);
+    this._service.addListener("resource-update", this.update.bind(this));
+
+    ActionHandlerInterface.apply(this);
```
chriskr: This doesn't really belong in `init`, it's more extending the prototype.

chriskr:
Partially not completely trivial code, needs good testing.

src/network/network_view.js
danfooo: Fixed the name, it wasn't actually making a function. Also that code is just moved.

src/build-application/build_resource_manager_1_0.js
chriskr: The just NetworkLogger sounds much better.

src/network/network_service.js
chriskr: Otherwise you would do `this._logger.get_request_context(NetworkLogger.CONTEXT_TYPE_CRAFTER)`.

src/network/network_service.js
chriskr: Ok, my bad.

hzr:
Looks fine!

src/build-application/build_resource_manager_1_0.js
danfooo: Good, fixing.

PR #51, opened 2012-08-22
Fixes for Release task 13, dfl 3295 network details overhaul

danfooo:


src/network/network_details_templates.js
```diff
@@ -377,12 +377,17 @@ templates._request_body = function(req, do_raw)
 templates._response_body = function(resp, do_raw, is_last_response)
 {
   var ret = [];
-
   var classname = "";
-  if ((resp.saw_responsefinished && resp.no_used_mimetype) ||
-      !resp.responsebody && resp.is_unloaded)
+
+  var should_track_content = (
```
hzr: Drop the parens IMO.

src/network/network_service.js
```diff
@@ -975,20 +918,58 @@ cls.NetworkLoggerEntryPrototype = function()
     this.events.push(evt);
   };
 
-  this.check_to_request_body = function(service)
+  this.check_to_get_body = function()
   {
-    // Decide if body should be fetched, for when content-tracking is off or it's a cached request.
-    if (
+    var should_get_body = (
```
hzr: Would also drop parens here.

src/network/network_service.js
```diff
@@ -975,20 +918,58 @@ cls.NetworkLoggerEntryPrototype = function()
...
+    // We could also avoid it when this.is_unloaded, but seems there it will 
+    // just be unsuccessful and we handle that.
+
+    if (should_get_body)
+    {
+      // Decide if body should be fetched, for when content-tracking is off or it's a cached request.
+      this.called_get_body = true;
+      if (this._current_response)
+        this._current_response.update_called_get_body(true);
+
```
hzr: Can remove this blank line.

src/network/network_service.js
```diff
@@ -975,20 +918,58 @@ cls.NetworkLoggerEntryPrototype = function()
...
+      var contentmode = cls.ResourceUtil.mime_to_content_mode(this.mime);
```
hzr: Nitpick: `contentmode` and `mime_to_content_mode`. Could use `content_mode` instead.

src/network/network_service.js
```diff
@@ -975,20 +918,58 @@ cls.NetworkLoggerEntryPrototype = function()
...
+      var typecode = {datauri: 3, text: 1}[contentmode];
```
hzr: Would use string property names and put them in order, so: `var typecode = {"text": 1, "datauri": 3}[contentmode];`

src/network/network_service.js
```diff
@@ -975,20 +918,58 @@ cls.NetworkLoggerEntryPrototype = function()
...
+      var tag = window.tagManager.set_callback(this, this._handle_get_resource);
```
hzr: `tag_manager`

src/network/network_service.js
```diff
@@ -1152,6 +1129,16 @@ cls.NetworkLoggerResponsePrototype = function()
...
+
+  this.update_called_get_body = function(called_get_body)
+  {
+    this.logger_entry_called_get_body = called_get_body;
+  };
+
+  this.update_get_body_unsuccessful = function(get_body_unsuccessful)
+  {
+    this.logger_entry_get_body_unsuccessful = get_body_unsuccessful;
+  };
```
hzr: Are the above methods really needed? Why not just set the property directly?

src/network/network_details_templates.js
danfooo: I thought of semicolon insertion, I don't know the rules well enough there. With a return statement this would be a very silent error, so maybe reason enough to keep so there's no doubt when reading it?

src/network/network_service.js
danfooo: Guess it's `type_code` too then. Fixing.

src/network/network_details_templates.js
hzr: There's no ASI issue here. `return` is one of the few cases where it's an issue. Just drop them IMO. :)

src/network/network_service.js
danfooo: I never changed things directly on `NetworkLoggerRequest` or `NetworkLoggerResponse` objects, but that's mostly because events on the entry are just passed on to them. I tried to keep that pattern, but I think I can change to setting properties directly for `logger_entry_*` properties to make it a bit shorter. Fixing.

src/network/network_service.js
p01: `transport_type` might be more accurate.

https://homes.oslo.osa/rchlodnicki/docs/master/index.html?service=./protofiles/resourcemanager/ResourceManager.1.3.proto

danfooo: Good :)

hzr:
Good.

PR #60, opened 2012-09-04
DFL-3548: Header overrides not working with Next

danfooo:


src/network/network_global_options_view.js
```diff
@@ -97,16 +97,24 @@ cls.NetworkOptionsView = function(id, name, container_class, html, default_handl
...
-    this._service.requestClearHeaderOverrides(null, []);
+    var tag = window.tag_manager.set_callback(this, this._handle_clear_header_overrides, [headers_to_set]);
+    this._service.requestClearHeaderOverrides(tag, []);
+  };
+
+  this._handle_clear_header_overrides = function(status, data, headers_to_set)
+  {
+    if (headers_to_set)
+    {
+      var args = [headers_to_set.map(function(e) {return [e.name, e.value] })];
```
hzr, line 110: One space before "return". Also, maybe "header" (or "h") instead of "e"?

p01, line 110: Nitpicking, but no `;`after the `return` ?

danfooo:
Fixing these, but opening a new pull request so I can do the fix in the same commit. Was also just moved code, but it makes sense we improve code even if we just move it.

PR #61, opened 2012-09-04
DFL-3548: Header overrides not working with Next

danfooo:


danfooo:
See https://github.com/operasoftware/dragonfly/pull/60

PR #66, opened 2012-09-12
Task dfl 3295 network details overhaul

chriskr:


danfooo:
Yes, it's ready, thanks :)

src/network/network_details_templates.js
```diff
@@ -8,93 +8,75 @@ window.templates.network || (window.templates.network = {});
...
-      ],
-      entry.touched_network ? [] : this.did_not_touch_network(entry),
-      requests_responses
-    ]
-  );
-};
+  var response_summary = [];
+  if (entry.touched_network && responsecode)
+  {
+    response_summary = [
```
hzr, line 74: "responsecode" (no underscore) and "responsesummary". Could use "response_code".

src/network/network_details_templates.js
```diff
@@ -8,93 +8,75 @@ window.templates.network || (window.templates.network = {});
...
+    ];
+  }
+  else
+  {
+    var local_url_types = [
+      cls.ResourceManager["1.2"].UrlLoad.URLType.FILE,
+      cls.ResourceManager["1.2"].UrlLoad.URLType.DATA
+    ];
+    response_summary = [
+      "p", local_url_types.contains(entry.urltype) ? ui_strings.S_NETWORK_NOT_REQUESTED
```
hzr: Content on the next line.

src/network/network_details_templates.js
```diff
@@ -8,93 +8,75 @@ window.templates.network || (window.templates.network = {});
...
+  }
 
-templates.did_not_touch_network = function(entry)
-{
-  var local_url_types = [
-    cls.ResourceManager["1.2"].UrlLoad.URLType.FILE,
-    cls.ResourceManager["1.2"].UrlLoad.URLType.DATA
+  return [
+    ["p",
+      [
```
hzr: We don't really have a standard for it, but this is not how we usually write templates. I'd do

``` javascript
["span",
  (entry.current_request && entry.current_request.method + " ") || ""
]
```

Same in other places.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+    cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART
+  ];
+  var attrs = ["class", cls.HTTPHeaderTokenizer.classnames[token[TYPE]]];
 
-  if (token[TYPE] === cls.HTTPHeaderTokenizer.types.NAME)
-  {
-    attrs.extend(["data-spec", "http#" + token[STR].trim()]);
-  }
-  else if (token[TYPE] === cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART)
+  if (highlighted_types.indexOf(token[TYPE]) != -1)
```
hzr: We have `contains()`.

src/network/network_details_templates.js
```diff
@@ -181,122 +217,130 @@ templates._request_headers = function(req, do_raw)
...
     if (resp.header_tokens.length)
     {
-      var data_spec_firstline_tokens = [1];
-      var state_holder = new cls.HTTPHeaderTokenizer.TokenStateholder(data_spec_firstline_tokens);
-      var map_func = this._header_token_templ.bind(this, state_holder);
-      return [
-        ["h2", ui_strings.S_NETWORK_REQUEST_DETAIL_RESPONSE_TITLE],
-        this._pre(resp.header_tokens.map(map_func))
-      ];
+      var RESPONSECODE = 1;
```
hzr: If you change to "response_code", this should change too.

src/network/network_details_templates.js
```diff
@@ -181,122 +217,130 @@ templates._request_headers = function(req, do_raw)
...
-  }
 
-  ret.extend(this.headers_list(resp.response_headers));
-  return ["tbody", ret.map(this._col_or_row)];
+    if (resp.firstline_tokens.length)
+    {
+      ret.unshift(this.headers_tonkenized(resp.firstline_tokens, [RESPONSECODE]));
+    }
+  }
+  if (!ret)
```
hzr, line 317: Newline before.

src/network/network_details_templates.js
```diff
@@ -181,122 +217,130 @@ templates._request_headers = function(req, do_raw)
...
+      var tokenizer = new cls.HTTPHeaderTokenizer();
+      tokenizer.tokenize(resp.first_line, this._token_receiver.bind(this, resp.firstline_tokens))
     }
-    ret.push(head);
-  }
 
-  ret.extend(this.headers_list(resp.response_headers));
-  return ["tbody", ret.map(this._col_or_row)];
+    if (resp.firstline_tokens.length)
+    {
```
hzr: Can skip braces.

src/network/network_style.css
```diff
@@ -304,55 +297,87 @@
...
+}
+
+.network-detail-overlay .wrap img
+{
+  max-width: 100%;
+}
+
+.network-detail-overlay p
+{
+  font: 11px "menu";
```
hzr: If what you want is the font used for menus, this is not valid. I guess you want:

``` css
font: menu;
font-size: 11px;
```

src/network/network_style.css
```diff
@@ -304,55 +297,87 @@
...
+  margin: 0 0 0 -5px; /* todo: 10px 0 10px -5px; looks better to me but is inconsistent with dom panel headers */
 }
 
-.network-details-container h2,
 .request-crafter h2,
 .network-options h2
 {
   font-size: 125%;
-  margin: 15px 0 10px 0;
+  padding: 10px 0 5px 0;
```
hzr: `10px 0 5px`

src/network/network_view.js
```diff
@@ -319,66 +290,17 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
-    {
-      var selected_node = document.querySelector("[data-object-id='" + this._selected + "']");
+      var selected_node = document.querySelector("[data-object-id='" + this.selected + "']");
       if (selected_node &&
           selected_node[accessor] &&
           selected_node[accessor].dataset.objectId)
       {
-        this._selected = selected_node[accessor].dataset.objectId;
-        selected_node = document.querySelector("[data-object-id='" + this._selected + "']");
+        this.selected = selected_node[accessor].dataset.objectId;
```
hzr, line 302: "selected_id" maybe? At least something clearer.

src/network/network_view.js
```diff
@@ -440,20 +357,12 @@ cls.NetworkLogView = function(id, name, container_class, html, default_handler,
...
-    if (evt.target.hasClass("entry-details"))
-    {
-      this._details_scroll_top = evt.target.scrollTop;
-      this._details_scroll_left = evt.target.scrollLeft;
-    }
-    else
-    {
-      this._container_scroll_top = target.firstChild.scrollTop;
-    }
+    this._container_scroll_top = target.firstChild.scrollTop;
```
hzr: firstElementChild might be safer, unless it's guaranteed to be an element.

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+      {
+        entry.check_to_get_body();
+        container.clearAndRender(this._render_details_view(entry));
+        this.text_search.update_search();
+        if (this._details_scroll_top)
+          container.scrollTop = this._details_scroll_top;
+
+        if (this._details_scroll_left)
+          container.scrollLeft = this._details_scroll_left;
+
```
hzr: Extra blank line.

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+  };
+
+  this._render_details_view = function(entry)
+  {
+    return templates.network.details(entry);
+  };
+
+  this._on_toggle_expand_request_response = function(event)
+  {
+    var KEY = event.target.dataset.isResponse ? "expand-responses" : "expand-requests";
```
hzr: I don't see why this necessarily has to be a constant.

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+        this.needs_instant_update = true;
+        this.update();
+        break;
+      }
+    }
+  };
+
+  this._init = function(id, container_class, html, default_handler)
+  {
+    messages.addListener("setting-changed", this._on_setting_changed.bind(this));
```
hzr, line 875: window.messages

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+
+    var eh = window.eventHandlers;
```
hzr: event_handlers

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+    eh.scroll["network-detail-overlay"] = this._on_scroll.bind(this);
+    eh.click["toggle-expand-request-response"] = this._on_toggle_expand_request_response.bind(this);
+
+    ActionHandlerInterface.apply(this);
+    this.handle = function(action_id, event, target)
+    {
+      var parent_view = window.views[this.parent_view_id];
+      if (parent_view)
+        parent_view.handle.apply(parent_view, arguments);
+
```
hzr: Extra blank line.

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+    ActionBroker.get_instance().register_handler(this);
+
+    this.init(id, container_class, html, default_handler);
+  }
+}
+
+cls.NetworkDetailOverlayViewPrototype.prototype = new OverlayView();
+cls.NetworkDetailOverlayView.prototype = new cls.NetworkDetailOverlayViewPrototype();
+
+
```
hzr: Extra blank line.

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+            title: ui_strings.S_SEARCH_INPUT_TOOLTIP,
+            label: ui_strings.S_INPUT_DEFAULT_TEXT_SEARCH
+          }
+        ]
+      }
+    ]
+  });
+
+  var text_search = window.views["network-detail-overlay"].text_search = new TextSearch();
+  window.eventHandlers.input["network-details-text-search"] = function(event, target)
```
hzr: event_handlers

src/network/network_view.js
```diff
@@ -878,8 +796,188 @@ cls.NetworkLog.create_ui_widgets = function()
...
+
+  var on_view_destroyed = function(msg)
+  {
+    if (msg.id == "network-detail-overlay")
+    {
+      text_search.cleanup();
+    }
+  };
+
+  messages.addListener("view-created", on_view_created);
```
hzr, line 984: `window.messages` and `add_listener`. The latter one should possibly be fixed in other places.

src/ui-scripts/celloverlay.js
```diff
@@ -0,0 +1,82 @@
...
+    this.parent_view_id = parent_view_id;
+    this.is_active = false;
+    window.messages.add_listener("show-view", this._onview_show.bind(this));
+    window.messages.add_listener("hide-view", this._onview_hide.bind(this));
+    var layout_rough =
+    {
+      dir: "h",
+      children:
+      [
+        { tabbar: { tabs: [], is_hidden: true, is_empty: true } , width: 150, min_width: 7, name: view_id },
```
hzr, line 21: Not your code I believe, but we don't usually have whitespace before the first and after the last value in an object literal.

src/ui-scripts/cells.js
```diff
@@ -499,7 +564,7 @@
   {
     var dim = this.dir == VER ? 'height' : 'width';
     var max = this[dim];
-    var child = null, i = 0, sum = 0, length = this.children.length, temp = 0;
+    var child = null, i = 0, sum = 0, length = this.children.length, temp = 0, min = 0;
```
hzr: Not your code either, but we should have one var per line.

src/ui-scripts/ui-actions.js
```diff
@@ -42,7 +42,7 @@ var EventHandler = function(type, is_capturing, handler_key, cancel_bubble)
     while (ele)
     {
       handler = ele.getAttribute(handler_key);
-      while (!(handler && eventHandlers[type][handler]) && (ele = ele.parentNode))
+      while (!(handler && window.eventHandlers[type][handler]) && (ele = ele.parentNode))
       {
```
hzr, line 46: event_handlers

src/network/network_details_templates.js
```diff
@@ -8,93 +8,75 @@ window.templates.network || (window.templates.network = {});
...
+      "class", (entry.error_in_current_response ? templates.ERROR_RESPONSE
+                                                : "") + " response-summary"
+    ];
+  }
+  else
+  {
+    var local_url_types = [
+      cls.ResourceManager["1.2"].UrlLoad.URLType.FILE,
+      cls.ResourceManager["1.2"].UrlLoad.URLType.DATA
+    ];
```
chriskr: Looks like it would make sense to make `local_url_types` an outer scope variable (or constant, actually).

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
   return [
-    templates._request_headers(request, do_raw),
-    templates._request_body(request, do_raw)
+    "div", [
+      show_header ? templates._headline(false, expanded) : [],
+      ["div",
+        [
+          show_headers ? templates._request_headers(request, do_raw) : [],
+          show_body ? templates._request_body(request, do_raw) : []
+        ], "class", "foldable"
```
chriskr: You don't need here the outer brackets AFAICS.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+      ],
+      "class", (expanded ? "unfolded" : "")
+    ]
```
chriskr: Also this bracket level is not needed AFAICS.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+      show_header ? this._headline(true, expanded) : [],
+      ["div",
+        [
+          show_headers ? this._response_headers(response, do_raw) : [],
+          show_body ? this._response_body(response, do_raw, is_last) : []
+        ], "class", "foldable"
+      ],
+      "class", (expanded ? "unfolded" : "")
+    ]
   ];
```
chriskr, line 165: The same here, you can remove some bracket layers.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+    "div",
+      ["input",
+        "type", "button",
+        "class", is_unfolded ? "unfolded" : ""
+      ],
+      is_response ? ui_strings.S_NETWORK_REQUEST_DETAIL_RESPONSE_TITLE
+                  : ui_strings.S_NETWORK_REQUEST_DETAIL_REQUEST_TITLE,
+      "handler", "toggle-expand-request-response",
+      "class", "header"
+  ].concat(is_response ? ["data-is-response", "data-is-response"] : []);
```
chriskr: I have done such thing previously too, but i think it's cleaner to have a proper if statement and then just push to the template what you have to push.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+
 templates._header_token_templ = function(state, token)
 {
   var TYPE = 0;
   var STR = 1;
-  var attrs = ["class", "header-token-type-" + cls.HTTPHeaderTokenizer.classnames[token[TYPE]]];
+  var highlighted_types = [
+    cls.HTTPHeaderTokenizer.types.NAME,
+    cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART
+  ];
```
chriskr, line 180: Looks like this could be in an outer scope.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
-  }
-  else if (token[TYPE] === cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART)
+  if (highlighted_types.indexOf(token[TYPE]) != -1)
   {
-    if (state.data_spec_firstline_tokens.contains(state.firstline_tokens))
+    if (token[TYPE] === cls.HTTPHeaderTokenizer.types.NAME)
     {
-      // Add data-spec attributes on certain firstline tokens, tracked in state
-      attrs.extend(["data-spec", "http#" + (token[STR]).trim()]);
+      attrs.extend(["data-spec", "http#" + token[STR].trim()]);
```
chriskr: Or just `attrs.push("data-spec", "http#" + token[STR].trim());`.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+    }
+    else if (token[TYPE] === cls.HTTPHeaderTokenizer.types.FIRST_LINE_PART)
+    {
+      if (state.data_spec_firstline_tokens.contains(state.firstline_tokens))
+      {
+        // Add data-spec attributes on certain firstline tokens, tracked in state
+        attrs.extend(["data-spec", "http#" + (token[STR]).trim()]);
```
chriskr: As above, `push` seems more efficient here.

src/network/network_details_templates.js
```diff
@@ -122,46 +104,100 @@ templates._request = function(request, is_last, do_raw)
...
+      }
+      state.firstline_tokens++;
     }
-    state.firstline_tokens++;
+    var buffer = state.str_buffer;
+    var ret = [["span", token[STR]].concat(attrs)];
```
chriskr: With `attrs.unshift` you don't create an additional array.

Or actually why not just start the template with this?

src/network/network_details_templates.js
```diff
@@ -181,122 +217,130 @@ templates._request_headers = function(req, do_raw)
...
+    }
+  }
+  return templates._pre(ret);
 };
 
-templates._response_headers = function(resp, do_raw)
+templates._reduce_tokens = function(map_func, previous_val, current_val, index){
+  if (index == 1)
+    previous_val = map_func(previous_val);
+  return previous_val.concat(map_func(current_val));
```
chriskr: `extend` would be more efficient here.

src/network/network_service.js
```diff
@@ -1038,6 +1004,9 @@ cls.NetworkLoggerRequestPrototype = function()
   this._update_event_requestheader = function(event)
   {
     this.request_headers = event.headerList;
+    // Don't set a first_line for a SPDY request
+    if (event.raw && event.raw.indexOf("\n") != -1)
```
chriskr: You can use here `contains` too.

src/network/network_service.js
```diff
@@ -1104,6 +1073,9 @@ cls.NetworkLoggerResponsePrototype = function()
     // At the time of the this event, it's possible that more than the header
     // has been read from the socket already.
     this.response_headers_raw = event.raw.split("\r\n\r\n")[0];
+    // Don't set a first_line for a SPDY request
+    if (this.response_headers_raw && this.response_headers_raw.indexOf("\n") != -1)
+      this.first_line = this.response_headers_raw.split("\n")[0];
```
chriskr, line 1078: `contains` works here too.

src/network/network_style.css
chriskr: Why does it need that font declaration at all? Should it not just inherit?

src/network/network_view.js
chriskr, line 302: Also this is element access without check if it exists. So far we have not used `dataset`. Is it support in ci 168?

src/network/network_view.js
chriskr: Also is element access without check if it exists.

src/network/network_view.js
chriskr: I would suggest `parent_view.handle.call(parent_view, action_id, event, target);` here. `arguments` is basically deprecated.

src/ui-scripts/celloverlay.js
chriskr, line 21: Copied that from client.js, all similar declarations have the same white spaces.

src/network/network_view.js
hzr: I'd agree about using call() here, but only because it looks nicer IMO. How is `arguments` "basically deprecated"? Are you referring to performance?

src/ui-scripts/cells.js
chriskr: I will fix this.

src/ui-scripts/ui-actions.js
chriskr, line 46: I will fix this.

src/network/network_view.js
chriskr: Yes. But i also think `call` makes more sense here.

src/network/network_details_templates.js
p01: Maybe even a dictionary/hashmap in cls.ResourceUtil ?

danfooo: Going for LOCAL_URL_TYPES within window.templates.network now. We have TEXT_TYPES there, they seem similar.

src/network/network_details_templates.js
p01: Or get rid of the `.concat( /* ... */ )` and shove the test and its `:[]`fallback directly in the array

src/network/network_details_templates.js
danfooo: Like

``` javascript
      "p", LOCAL_URL_TYPES.contains(entry.urltype) ?
        ui_strings.S_NETWORK_NOT_REQUESTED : ui_strings.S_NETWORK_SERVED_FROM_CACHE,
```

?

hzr: No, I'd move the whole conditional down.

src/network/network_details_templates.js
danfooo: We should have a standard for it. I think I'm changing how I do it myself too often, we would save some time with deciding. There could be different cases or two styles to choose on depending and what you do.

Agree in this case, not much use for a line just for `[`.

src/network/network_details_templates.js
danfooo: True. I will get used to that at _some_ point.

src/network/network_details_templates.js
danfooo, line 165: Good, fixing.

src/network/network_details_templates.js
danfooo: Was also wondering about a good value here, when you access over dataset the key is actually isResponse, so the value could be that. Would want it close to a boolean attribute as I can have it..

danfooo: But yes, fixing.

src/network/network_details_templates.js
p01: +1 starting the template with this.

src/network/network_details_templates.js
danfooo: But that's what we typically use extend for, no? For what else would we have extend then?

src/network/network_details_templates.js
p01: Here's the boolean `element.hasAttribute('data-is-response')` ;)

src/network/network_details_templates.js
```diff
@@ -181,122 +217,130 @@ templates._request_headers = function(req, do_raw)
...
   if (!req.request_headers)
-    ret.push(ui_strings.S_NETWORK_REQUEST_NO_HEADERS_LABEL);
+  {
+    ret.push(["span", ui_strings.S_NETWORK_REQUEST_NO_HEADERS_LABEL,
+              "class", templates.UI_CLASSNAME]);
+  }
   else
-    ret.extend(this.headers_list(req.request_headers));
+  {
+    ret.extend(req.request_headers.map(this._headers_pseudo_raw));
```
p01, line 217: isn't a regular `push` OK ?

src/network/network_details_templates.js
danfooo: Because then I can't do that conditionally. Ah, or I could start with "" if I don't have a buffer? I'm not sure what you mean with `attrs.unshift`, the outer array is so that I can put the buffer before the `span`. Was also thinking I could just make a separate return that would include the buffer, but I also need to clear it afterwards, that was much worse.

danfooo: Seems `[buffer, ["span", token[STR]].concat(attrs)];` is not working as a template when buffer is "".

src/network/network_details_templates.js
```diff
@@ -351,26 +412,24 @@ templates._request_body = function(req, do_raw)
...
         }
       }
     }
   }
-
-  if (do_raw)
-    return ret;
-  else
-    return ["tbody", ret.map(this._col_or_row)];
+  return [this._pre("\n"), templates._pre(ret)];
```
p01: What is the raison d'être of the `this._pre("\n")`?

src/network/network_details_templates.js
danfooo: Alright, pushing more then.

src/network/network_details_templates.js
danfooo: (Changed that whole map/reduce procedure completely now)

src/network/network_details_templates.js
danfooo, line 217: That map function returns an array which I want to "append" to `ret`.

src/network/network_details_templates.js
danfooo, line 317: We should define this, so I can learn it. I guess here it's to show that the two if blocks have no relation, will do that.

src/network/network_details_templates.js
danfooo: That's the body separator, adding `"\n"` as a constant.

src/network/network_style.css
danfooo: Legacy reasons. Removing.

src/network/network_view.js
danfooo, line 302: Would rather not change this (now), not worth it, adds more merging horror downwards. Has quite a few references and the APIs are changing.

On dataset, I thought I had seen it being used, but I don't see it now either. But ci 168 is the requirement for the host, not really for the client? We have it since 11.10, I think 3 O Desktop versions back should be okay to rely on.

src/network/network_view.js
danfooo: But we call it on `parent_view` anyway? What's the difference with just calling `parent_view.handle(action_id, event, target);`?

src/network/network_view.js
danfooo, line 984: Would like to postpone changing to addListener if that's okay. Lots and lots of changes to network_view.js in branches I'll merge this to.

src/network/network_view.js
chriskr, line 302: And on accessing element without check?

src/network/network_view.js
chriskr: `parent_view.handle(action_id, event, target);` is even better.

src/network/network_details_templates.js
chriskr: E.g. to extend an existing array with a return value of map.

src/network/network_view.js
danfooo, line 302: What do you mean? Think it's all checked for.

src/network/network_details_templates.js
danfooo: Okay, understand better now!

src/network/network_view.js
chriskr, line 302: Sorry, i somehow overlooked that.

danfooo:
Three new commits, can you take a look? Were we done apart from those?

hzr:
All fine.

p01:
Good!

PR #69, opened 2012-09-14
DFL-3589 Implement extension name handling using info provided by CORE-41612

rchl:


src/ecma-debugger/runtimes.js
```diff
@@ -38,7 +40,7 @@ cls.EcmascriptDebugger["6.0"].ExtensionRuntime = function(rt)
   this.type = "extension";
   this.id = rt.runtime_id;
   this.uri = rt.uri;
-  this.title = "Extension Runtime " + rt.runtime_id;
+  this.title = "Extension" + (rt.extension_name ? ': ' + rt.extension_name : '');
 };
```
chriskr, line 44: This should still use the old notation in case we don't get an extension name, e.g. the code should support older versions too. Also for consistency use double quotes.

chriskr:
Will this already address the issue that we don't handle script with the same name in different extensions?

src/ecma-debugger/runtimes.js
rchl, line 44: It still supports old clients. Just that it changes how info is displayed for them. I thought that there is no purpose in showing extension id here, whether it be old or new client.

chriskr, line 44: But then all extensions get exactly the same name for older versions? That is IMO not better. The id was used so that they are at least different.

p01, line 44: So you're suggesting :

```
 this.title = "Extension" + (rt.extension_name ? ': ' + rt.extension_name : ' Runtime '+ rt.runtime_id);
```

rchl, line 44: I don't necessarily agree that having randomly changing id in here helps user with anything but if you think so.

Having 'Runtime' in one case and not the other is a little bit inconsistent also.

danfooo, line 44: I agree with keeping the old behaviour when we don't have the name.

chriskr, line 44: I'm not claiming this is a good solution for older versions, but i think it's better workaround than to show always exactly the same name. Then the user has at least a chance to differentiate them.

rchl:
Chris, you meant the CORE-46310 issue? If yes then I guess we can use extension name as a fairly reliable way of identifying the script.

chriskr:
Rafal, yes. OK, good.

p01:
+1

p01:
I suppose we'll merge this into the next bugfix branch.

PR #70, opened 2012-09-17
Finishing Network details

danfooo:


hzr:
Looks fine.

src/ui-strings/ui_strings-en.js
```diff
@@ -1263,8 +1263,8 @@ ui_strings.S_NETWORK_NOT_REQUESTED = "No request made.";
 /* DESC: String used when there is a request body we can't show the contents of directly. */
 ui_strings.S_NETWORK_N_BYTE_BODY = "Request body of %s bytes";
 
-/* DESC: Name of networks raw-view setting, shows requests and responses raw instead of parsed */
-ui_strings.S_NETWORK_RAW_VIEW_LABEL = "Show raw requests and responses";
+/* DESC: Name of networks raw-view setting, shows requests and responses parsed as opposed to raw */
+ui_strings.S_NETWORK_RAW_VIEW_LABEL = "Show parsed requests and responses";
```
chriskr: Should this not be "Show parsed requests and response headers"? The response is not actually parsed really, that would be a DOM.

p01: That string actually applies to the `view-parsed` setting.
Other than this "incomplete" switch between **raw** and **parsed**, it looks fine.

danfooo: It applies to the request headers and body, and response headers. So this is a simplification. Maybe `Show requests and responses in parsed mode`, to reflect that this is modifies how _and_ what is shown?

danfooo: By incomplete you mean that the icon is missing? That will be fixed with Tonys pull request. Have fixed the name of the ui-string.

PR #71, opened 2012-09-17
Network Details Style tweaks.

TonyNedyalkov:


src/network/network_style.css
```diff
@@ -339,23 +356,38 @@
   white-space: nowrap;
 }
 
+.network-detail-overlay th .not-content
```
danfooo: I think this should not be limited to `th`, .not-content is also used for some messages, like missing headers, or when we can't display data of a certain type. I should probably add a test, but you can see it when you follow the steps in https://bugs.opera.com/browse/DFL-3590

src/network/network_style.css
```diff
@@ -339,23 +356,38 @@
...
+{
+  padding: 0px;
```
danfooo: Should be `0`, in other places too.

src/network/network_style.css
```diff
@@ -339,23 +356,38 @@
...
   border: none;
-  border-top: 1px solid #a4a4a4;
+  border-top: 1px solid #ccc;
   margin: 7px 0 -8px 0;
 }
 
-.network-detail-overlay div.header
-{
-  margin: 0 0 0 -5px; /* todo: 10px 0 10px -5px; looks better to me but is inconsistent with dom panel headers */
+.response-summary + div .header {
```
danfooo: `{` should be on the next line.

src/network/network_style.css
```diff
@@ -339,23 +356,38 @@
...
 }
 
+.network-detail-overlay th .not-content
+{
+  padding: 0px;
+  font-style: normal;
+  border-collapse: collapse;
+  color: #808080;
+  cursor: default;
+  display: block;
```
chriskr: `display: block` in conjunction with `border-collapse: collapse;` makes limited sense?

src/network/network_view.js
```diff
@@ -909,7 +909,7 @@ cls.NetworkDetailOverlayView.create_ui_widgets = function()
     },
     // key-label map
     {
-      "view-parsed": ui_strings.S_NETWORK_RAW_VIEW_LABEL,
+      "view-parsed": ui_strings.S_NETWORK_PARSED_VIEW_LABEL,
```
danfooo, line 912: (Pretty sure this just shows here because I initially broke it while merging)

src/ui-style/debugger_style.css
```diff
@@ -446,13 +441,13 @@ header input,
   transition: 0.2s;
 }
 
-.unfolded > header
+.unfolded .foldable
```
chriskr: This change looks a bit dangerous to me: have you taken all views which are using `header` in regard? Or does this only apply to the new view?

src/ui-style/ui.css
```diff
@@ -897,6 +897,7 @@ view-slider
 {
   display: block;
   position: absolute;
+  z-index: 2;
```
chriskr: Why do we need to add here a z-index?

src/ui-style/ui.css
```diff
@@ -1845,5 +1846,6 @@ info {
 {
   position: absolute;
   border: 1px solid #999;
-  box-shadow: 0 2px 5px hsla(0, 0%, 0%, 0.5);
+  box-shadow: 0 5px 5px hsla(0, 0%, 0%, 0.30),
+              0 2px 2px hsla(0, 0%, 0%, 0.25)
 }
```
chriskr, line 1850: What is this change about?

src/network/network_style.css
```diff
@@ -312,6 +327,11 @@
   max-width: 100%;
 }
 
+.network-detail-overlay p
```
hzr: Same styling as `.network-detail-overlay .foldable`, should they be merged?

danfooo: Yes, merged them and cleaned that up a bit.

danfooo:
Is this done? @chriskr I think you should look at the review fix

PR #75, opened 2012-09-19
Fix for most of DFL-3607: Minor styling issues in Network details

danfooo:


src/network/network_style.css
```diff
@@ -359,6 +359,8 @@
 .network-detail-overlay .not-content
 {
   padding: 0;
+  font: menu;
+  font-size: 11px;
```
chriskr: The style for ui labels should ideally only be defined once. I suggest to either add your selector to the according declaration in ui.css (at the very top) or to add to the selectors there a more generic class name for ui labels.

chriskr:
Other then that it looks good.

danfooo:
Fixed it, but weird having two classnames on a const, but seemed like the smallest solution.

hzr:
Looks good.

PR #76, opened 2012-09-20
Removed unused properties; DFL-3614: 'Name' and 'Value' labels in networ...

danfooo:
...k details should not have 'not-content' class


hzr:
Fine.

PR #77, opened 2012-09-20
DFL-3615: Line Wrap setting doesn't work if a table with long strings is...

danfooo:
... shown in the network details view


hzr:
Looks fine.

danfooo:
One more..

danfooo:
Cool, thanks :)

PR #84, opened 2012-09-27
DFL-2502 Don't handle horizontal scrolling and improve Mac scrolling spe...

chriskr:
...ed.

For mac we'll of course still scroll by lines but number of lines scrolled
should be equivalent to number of pixels that would otherwise be scrolled.

### DIFF
--- a/src/ecma-debugger/js-source-view.js
+++ b/src/ecma-debugger/js-source-view.js
@@ -1016,7 +1016,7 @@ cls.JsSourceView = function(id, name, container_class)
 
   window.eventHandlers.mousewheel['scroll-js-source-view'] = function(unit, event, target)
   {
-    if (event.wheelDeltaX !== undefined && event.wheelDeltaX != 0)
+    if (event.wheelDeltaX)
       // Horizontal scrolling is handled natively by the browser.
       return;
 
--- a/src/ecma-debugger/js-source-view.js
+++ b/src/ecma-debugger/js-source-view.js
@@ -988,7 +988,7 @@ cls.JsSourceView = function(id, name, container_class)
 
   this._get_lines_from_delta = function(delta, unit)
   {
-    var lines;
+    var lines = 0;
     if (unit == UNIT_LINES)
       lines = delta;
     else if (unit == UNIT_PIXELS)
@@ -1006,15 +1006,13 @@ cls.JsSourceView = function(id, name, container_class)
       {
         // Enough delta to scroll at least one line, round delta
         // to full integer towards 0 and store remainder for later.
-        lines = delta >= 1 ? Math.floor(delta) : Math.ceil(delta);
+        lines = delta > 0 ? Math.floor(delta) : Math.ceil(delta);
         _accumulated_delta = delta % 1;
       }
       else
         // Not enough delta accumulated to scroll.
         lines = 0;
     }
-    else
-      lines = 0;
 
     return lines;
   }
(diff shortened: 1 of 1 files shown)


hzr:
Looks fine.

src/ecma-debugger/js-source-view.js
```diff
@@ -981,10 +981,54 @@ cls.JsSourceView = function(id, name, container_class)
     }
   };
 
-  eventHandlers.mousewheel['scroll-js-source-view'] = function(event, target)
+  var _last_delta = 0;
+  var _accumulated_delta = 0;
+  const UNIT_LINES = 1;
```
danfooo: I think we shouldn't use `CONST` in new code.

src/ecma-debugger/js-source-view.js
```diff
@@ -981,10 +981,54 @@ cls.JsSourceView = function(id, name, container_class)
...
+        // Not enough delta accumulated to scroll.
+        lines = 0;
+    }
+    else
+      lines = 0;
+
+    return lines;
+  }
+
+  eventHandlers.mousewheel['scroll-js-source-view'] = function(unit, event, target)
```
hzr: Use `window.event_handlers` here.

src/ecma-debugger/js-source-view.js
rchl: Whynut? It's THE feature.

rchl: I mean 'future'.

p01: You know what's the future too ? The EDIT COMMENT feature. And the future is now ;)

src/ecma-debugger/js-source-view.js
```diff
@@ -981,10 +981,52 @@ cls.JsSourceView = function(id, name, container_class)
...
+        // Not enough delta accumulated to scroll.
+        lines = 0;
+    }
+
+    return lines;
+  }
+
+  eventHandlers.mousewheel['scroll-js-source-view'] = function(unit, event, target)
+  {
+    if (event.wheelDeltaX !== undefined && event.wheelDeltaX != 0)
```
chriskr: This seems (almost) the same as `if (event.wheelDeltaX)`?

rchl: I thought more explicit code will be somewhat more self-documenting but maybe that is not needed.

chriskr:
This looks now good.

p01:
Yup. All good!

src/ecma-debugger/js-source-view.js
```diff
@@ -981,10 +981,49 @@ cls.JsSourceView = function(id, name, container_class)
...
+  var _accumulated_delta = 0;
+  var UNIT_LINES = 1;
+  var UNIT_PIXELS = 2;
+
+  this._get_lines_from_delta = function(delta, unit)
   {
-    this._scroll_lines((event.detail > 0 ? 1 : -1) * 3 , event, target);
-  }.bind(this);
+    var lines = 0;
+    if (unit == UNIT_LINES)
```
chriskr, line 992: Only realized that now, but i think it would actually make more sense to do that branch check only once instead of in each scroll event, e.g. implement `_get_lines_from_delta` platform dependent. But only a detail.

p01, line 992: Agreed!

PR #86, opened 2012-09-27
Dfl 3440

chriskr:


hzr:
Looks good.

danfooo:
Is there any way we could have a detection and use the keep the old and the new handling side by side? Code looks fine.

chriskr:
There is already a detection and a fallback to the old behavior.

danfooo:
Ah, great.

PR #89, opened 2012-09-27
DFL-3538 Change pretty-printing icon

chriskr:


hzr:
Nice.

p01:
Would be even nicer if the lines had different length.
But it's already one notch better than the current icon.

chriskr:
We have tried different line lengths (and a lot of other things too), it actually doesn't look better.

danfooo:
I see an airplane. Maybe because I am on an airplane.

No, seriously, I don't like all our line-y icons. The views are lines of source already, repeating that in every icon doesn't make sense. It has no visual hold, nothing to get used to or remember.

This is okay since it fits our current icon style, but it's not good.

PR #94, opened 2012-10-01
Fixed #93, Opera throws an exception in 11.50.

### LINKED ISSUE #93: Dragonfly not working with opera 11.50 anymore
Hiya,
Not sure if dragonfly is supposed to support older versions of opera,
but fyi dragonfly doesn't work in opera 11.50 anymore (linux version).

Dragonfly loads forever, and error console reports css errors and a js syntax error (see below)

hzr:


danfooo:
Good.

hzr:
@danfooo @p01 There was more to fix actually.

PR #95, opened 2012-10-01
Made keyboard test work with shortcut map instead of static list

danfooo:


hzr:
Good.

chriskr:
Yes, looks good.

PR #96, opened 2012-10-02
Bugfix package 10

danfooo:


danfooo:
(Doesn't change anything in the build, only affects `test-scripts/`)

PR #100, opened 2012-10-02
Fix exception in 11.50

hzr:


hzr:
This should work in 11.50 now.

PR #107, opened 2012-10-08
Fix DFL-3650 - need to escape control characters before parsing with innerHTML

hzr:


src/syntaxhighlight/js/tokenizer.js
```diff
@@ -356,7 +356,40 @@ cls.SimpleJSParser.prototype = new function()
...
+    '\u0016': '\\u0016',
+    '\u0017': '\\u0017',
+    '\u0018': '\\u0018',
+    '\u0019': '\\u0019',
+    '\u001a': '\\u001a',
+    '\u001b': '\\u001b',
+    '\u001c': '\\u001c',
+    '\u001d': '\\u001d',
+    '\u001e': '\\u001e',
+    '\u001f': '\\u001f'
```
hzr: This would obviously be nicer with a replace function, but I followed the old format to mitigate the risk of messing up something deep down in the tokenizer.

danfooo:
Good!

src/syntaxhighlight/js/tokenizer.js
chriskr: `\u0009`, `\u000A` and  `\u000D` are valid characters in XML, they don't need to be escaped.

Has someone tried to run DFL in text/html? There are some unknown elements, but that shouldn't be really a problem technically?

Also this change would break the tooltip after such a char. Matching is done on the real source based on character count offset. An alternative approach would be to replace with a small markup snippet, e.g. for `\u0004` `<span data-control-char="EOT"> </span>` and display that with CSS similar to how e.g. Sublime Text displays control chars, a black box with the ASCI name (`content: attr(data-control-char);`.

(Actually that wouldn't work either, i think.)

hzr: About those characters, right, thanks!

I'm not sure if it's worth special-casing these for tooltips as they will appear very rarely. But I can think of some other way of doing it.

Also, back on vacation with you! ;)

hzr: Tried running it as text/html now. Works mostly fine, only found minor issues with some styling and no being able to resize panels (maybe because some parent element check fails now, I dunno).

hzr:
Fixed the tooltip issue. The spans contains the [control pictures](http://www.fileformat.info/info/unicode/block/control_pictures/list.htm) as "fallback" (not that this will ever be needed).

hzr:
And yes, this is totally ugly.

danfooo:
Should those control pictures be used in `escape_html` too?

hzr:
@danfooo I tried that, but that would require lots of changes to DOM editing. More than it's worth.

danfooo:
Ah, I see. Seems like a good enough solution then.

p01:
Looks better.

PR #109, opened 2012-10-11
Attempt for DFL-3654: Selectors (nth-child()) used for the sortable tabl...

danfooo:
...e are slow

### DIFF
--- a/src/console-logger/console_templates.js
+++ b/src/console-logger/console_templates.js
@@ -36,7 +36,7 @@ window.templates.errors.log_table = function(entries, allExpanded, expandedList,
   };
 
   return [
-    "table", entries.map(rowClosure),
+    "table", new window.helpers.EvenOddHandler().add_elems(entries.map(rowClosure)),
     "class", "sortable-table errors-table",
   ];
 };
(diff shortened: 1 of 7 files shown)


hzr:
Feels slightly over-engineered to me. Why not just generate the classes in the corresponding template?

p01:
I have to agree with @hzr, must be possible to add the corresponding classes in the templates. I do similar things in the templates of the new Resource Service

danfooo:
In https://github.com/operasoftware/dragonfly/pull/109/files#L4R179 for example, different rows come together from different templates. You'd pass `count`s between several map functions, and repeat the code that checks even/odd. Even when it's just one template function used with `map`, sometimes you return `[]`, so you couldn't work with the index that's passed. Okay, I admit that last one might be a theoretical problem.

@p01 if you do a similar thing, doesn't that mean it's useful to do it in one common place instead? I don't think it's expensive either.

Edit: Github kills my link here for some reason, I meant to point to a specific line `https://github.com/operasoftware/dragonfly/pull/109/files#L4R179`

PR #113, opened 2012-10-25
Profiler zooming

hzr:
Sending this to review now. It's still not completely done, but we might as well start reviewing.


src/profiler/profiler_style.css
```diff
@@ -2,33 +2,60 @@
 {
   position: absolute;
   left: 0;
-  top: 20px;
-  width: 160px;
-  border-top: 1px solid #e6e6e6;
+  top: 72px;
+  right: 0;
+  border-top: 1px solid rgba(0, 0, 0, 0.055);
```
p01, line 7: Why not `0.05` ?

0.05 × 255 ≈ 13
0.055 × 255 ≈ 14
0.06 × 255 ≈ 15

Additionally, further down there are some box-shadows with `.05` opacity. Not `.055`

hzr, line 7: You would have to ask some designer I guess. :) This comes from the tables originally. The box-shadows are not related. I agree that `0.05` would suffice though.

src/network/network_request_crafting_view.js
```diff
@@ -197,7 +197,7 @@ cls.RequestCraftingView = function(id, name, container_class, html, default_hand
     if (!urldata) { return; }
     var current = this._input.get_value();
     current = current.replace(/^(\w+? )(.*?)( .*)/, function(s, m1, m2, m3, all) {return m1 + urldata.path + " " + urldata.protocol + "/1.1" ; });
-    current = current.replace(/^Host: .*$?/m, "Host: " + urldata.host);
+    current = current.replace(/^Host: .*$/m, "Host: " + urldata.host);
```
danfooo, line 200: How come this is changed now? Does it now no longer match the "Host" line if it's the last one?

src/profiler/profiler.1.0.responses.getevents.js
```diff
@@ -252,6 +252,8 @@ cls.Profiler["1.0"].Area = function(arr, parent)
   this.y = arr[1];
   this.w = arr[2];
   this.h = arr[3];
+  this.ox = arr[4];
+  this.oy = arr[5];
```
danfooo: Isn't this generated? How come the 1.0 file has this now?

src/profiler/profiler_style.css
```diff
@@ -142,60 +175,63 @@
...
+  min-width: 3px;
+  height: 12px;
+  margin: 5px 0;
+  box-shadow: inset 0 2px rgba(255, 255, 255, 0.09),
+              inset 0 0 0 1px rgba(0, 0, 0, 0.07),
+              inset 0 -1px rgba(0, 0, 0, 0.05);
 }
 
-.profiler-event-interval
+.profiler-event.expandable:hover
```
danfooo: Possible to avoid the double class selector? The next ones too.

src/profiler/profiler_templates.js
```diff
@@ -83,17 +83,24 @@ var ProfilerTemplates = function()
...
-  var BAR_HEIGHT = profiler_timeline_row_decl
-                 ? (parseInt(profiler_timeline_row_decl.height) +
-                    parseInt(profiler_timeline_row_decl.paddingTop) +
-                    parseInt(profiler_timeline_row_decl.paddingBottom))
-                 : 0;
-  var TOOLTIP_WIDTH = profiler_event_tooltip_decl
-                    ? parseInt(profiler_event_tooltip_decl.width)
-                    : 0;
+  var EVENT_MIN_WIDTH = profiler_event_decl ? parseInt(profiler_event_decl.minWidth) : 1;
+  var EVENT_SMALL_MIN_WIDTH = profiler_event_decl ? parseInt(profiler_event_small_decl.minWidth) : 1;
```
danfooo: Should this check on `profiler_event_small_decl` instead? Line 95 too.

src/profiler/profiler_templates.js
```diff
@@ -83,17 +83,24 @@ var ProfilerTemplates = function()
...
+  var EVENT_HEIGHT = profiler_timeline_row_decl
+                   ? (parseInt(style_sheets.getDeclaration(".profiler-timeline-row").height) +
+                      parseInt(style_sheets.getDeclaration(".profiler-timeline-row").paddingTop) +
+                      parseInt(style_sheets.getDeclaration(".profiler-timeline-row").paddingBottom))
```
danfooo: Seems `getDeclaration` doesn't cache, maybe should just get it once? Maybe I just don't get it, why does it depend on another declaration that you already have?

src/network/network_request_crafting_view.js
hzr, line 200: It changed because it's a syntax error according to the spec and broke Firefox and Chrome. But I think it was an accident that this was merged to this branch.
I don't think this breaks anything in Opera (or does it?), so we can probably just keep it.

src/profiler/profiler.1.0.responses.getevents.js
hzr: It is generated normally, but I added this manually. We should switch to the generated version, but it will most likely be the exact same thing, only with "1.1" in the file name.

src/profiler/profiler_templates.js
hzr: Indeed, thanks.

src/profiler/profiler_templates.js
hzr: Oops, fixed. (It is the same declaration.)

src/profiler/profiler_templates.js
```diff
@@ -211,69 +192,111 @@ var ProfilerTemplates = function()
...
+    var width = Math.round((event.interval.end - event.interval.start) * ms_unit);
+    var left = Math.round((event.interval.start - interval_start) * ms_unit);
+    var column = this._order.indexOf(event.type);
+    return (
+      ["div",
+       "style",
+         "width: " + width + "px; " +
+         "left: " + left + "px; " +
+         "top: " + (column * EVENT_SMALL_HEIGHT) + "px;",
+       "class", "profiler-event-small event-type-" + event.type + "-selftime" // not actually selftime, just using that color
```
danfooo, line 230: Line too long for github

src/profiler/profiler_templates.js
p01: Correct, why not use `profiler_timeline_row_dec` to access the `height`, `paddingTop`and `paddingBottom`

src/profiler/profiler_templates.js
```diff
@@ -211,69 +192,111 @@ var ProfilerTemplates = function()
...
-                           "linear-gradient(90deg, " + color + " 0, " +
-                                            color + " " + self_time + "px, " +
-                                           "transparent " + self_time + "px);",
-       "class", "profiler-event profiler-event-interval event-type-" + event.type +
-                (event.eventID == selected_id ? " selected" : "") +
-                (is_expandable ? " expandable" : " non-expandable"),
+         "width: " + width + "px; " +
+         "left: " + left + "px; " +
+         "top: " + (column * EVENT_HEIGHT) + "px; " +
+         (width > 46340 ? ("background-size: " + Math.floor(46340 / width * 100) + "%; ") : "") + // workaround for CORE-48579
```
danfooo, line 252: Line too long

src/profiler/profiler_templates.js
```diff
@@ -83,17 +83,24 @@ var ProfilerTemplates = function()
...
+  var EVENT_HEIGHT = profiler_timeline_row_decl
+                   ? (parseInt(style_sheets.getDeclaration(".profiler-timeline-row").height) +
+                      parseInt(style_sheets.getDeclaration(".profiler-timeline-row").paddingTop) +
+                      parseInt(style_sheets.getDeclaration(".profiler-timeline-row").paddingBottom))
+                   : 1;
+  var EVENT_SMALL_HEIGHT = profiler_event_decl ? parseInt(profiler_event_small_decl.height) : 1;
+
+  var MIN_DURATION = ProfilerView.MIN_DURATION;
+
+  var HAS_UNPREFIXED_GRADIENTS = (function() {
```
danfooo, line 99: Would be nice to have that available globally. Network just puts both gradients for example, would be good to do it similar.

src/profiler/profiler_templates.js
```diff
@@ -317,7 +334,8 @@ var ProfilerTemplates = function()
                    ui_strings.S_PROFILER_START_TIME + ": ",
                  "class", "profiler-event-tooltip-label"
                 ],
-                this.format_time(event.interval.start)
+                this.format_time(event.interval.start) +
+                " (rel: " + this.format_time(event.interval.start - range_start) + ")"
```
danfooo: What is "rel"?

src/profiler/profiler_templates.js
hzr, line 99: Yea, if you want this in Network, move it somewhere appropriate. But I think that this is much less important in Network. I did this to save some time on CSS parsing which became heavy when moving the zoomer overlay a lot. I don't think Network suffers from this.

src/profiler/profiler_templates.js
hzr: Relative time from the start, i.e. if the event start time is at 150 ms and your zoom area starts at 100 ms, the relative start time is 50 ms.
And yea, it kinda sucks (I wanted something shorter than "relative" because it was usually too long for the tooltip min-width). It also didn't make it to the string review.

src/profiler/profiler_view.js
```diff
@@ -165,7 +132,15 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
         return this[timeline_mode.id];
       }, this);
       if (got_all_responses)
+      {
+        if (this._timeline_list && this._timeline_list.interval)
+        {
+          // TODO: should really be set according to the visible events
```
danfooo, line 138: Would it then "snap" to actual events?

src/profiler/profiler_templates.js
p01: We could use the Mathematical delta symbol.
http://www.fileformat.info/info/unicode/char/1d6ab/index.htm

hzr: @p01, yea, that might be an idea, thanks!

src/profiler/profiler_view.js
hzr, line 138: Not sure what you mean. The zoom area would "snap" to the visible events when filtering, yes.

src/profiler/profiler_view.js
```diff
@@ -181,39 +156,75 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+    var timeline_list = this._timeline_list;
+    var interval = timeline_list.interval;
+    var event_list = timeline_list.eventList.filter(function(event) {
+      return event.time > this._min_event_time;
+    }, this);
+    var aggregated_event_list = this._get_aggregated_event_list(event_list);
+    var width = this._container.clientWidth - AGGREGATED_EVENTS_WIDTH;
+
+    // TODO: Check if these are already appended
+    var frag = document.createDocumentFragment();
```
danfooo, line 177: To `frag`? Looks like they aren't :P

src/profiler/profiler_view.js
```diff
@@ -181,39 +156,75 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+
+    container.innerHTML = "";
+    container.appendChild(frag);
+
+    this._timeline_width = width;
+    this._zoomer.set_zoomer_element(this._zoomer_ele);
+    if (this._set_zoomer)
+      this._zoomer.set_current_area();
+
+    // TODO: if _set_zoomer == false, reset()
```
danfooo: Is that needed?

src/profiler/profiler_view.js
```diff
@@ -181,39 +156,75 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+    this._timeline_ele.setAttribute("handler", "profiler-event");
+    this._timeline_times_ele = document.createElement("div");
+    this._timeline_times_ele.className = "profiler-timeline-background";
+
+    if (this._timeline_list && this._timeline_list.eventList && this._timeline_list.eventList.length)
+    {
+      this._update_view();
+    }
+    else
+    {
```
danfooo, line 219: Probably none of the elements created above are needed when we end up here, maybe move those creations into the `if` part?

src/profiler/profiler_view.js
```diff
@@ -261,54 +269,59 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+      };
+    });
+    event_list.forEach(function(event) {
+      var type = event.type;
+      for (var i = 0, item; item = list[i]; i++)
+      {
+        if (item.type == type)
+        {
+          item.time += event.time;
+          continue;
```
danfooo: There's not much more to skip?

src/profiler/profiler_view.js
hzr, line 219: They are not needed when we end up here, no, but they are needed when _update_view() is called. There is no reason to re-create them on every _update_view() though, which is why I only create them on createView().

src/profiler/profiler_view.js
hzr: Thanks, removed.

src/profiler/profiler_view.js
```diff
@@ -319,19 +332,10 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
-    return this._has_zero_at_first_event
-         ? (this._timeline_list &&
-            this._timeline_list.eventList &&
-            this._timeline_list.eventList[0] &&
-            this._timeline_list.eventList[0].interval.start)
-         : 0;
+    var button_title = this._profiler.is_active
+                     ? ui_strings.S_BUTTON_STOP_PROFILER
+                     : ui_strings.S_BUTTON_START_PROFILER;
+    window.toolbars[this.id].set_button_title("profiler-start-stop", button_title);
```
danfooo, line 337: I think we talked briefly about it, but is it really consistent to have an alternating text here? I think we have that nowhere else? We can't change the string for a while now, so it's probably the best option anyway.

hzr, line 337: Yea, I think it was OK here, because it's not a switch. Anything else would just be confusing. Having an active icon would make a lot of sense though.

src/ui-scripts/zoomer/zoomer.css
```diff
@@ -0,0 +1,52 @@
...
+    display: none;
+}
+
+.zoomer-overlay-handle
+{
+    position: absolute;
+    top: 0;
+    width: 1px;
+    height: 100%;
+    cursor: ew-resize;
```
danfooo, line 30: I think it would be nice if these became visible on `:hover`, so they would overlay border-left or border-right of `.zoomer-overlay`. Or maybe style border-left and -right of `.zoomer-overlay` different by default too, since those are interactive?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
+"use strict";
+
+// TODO: when width is 1px and mousewheeling, don't move
```
danfooo: Why?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+    this._update_bound = this._update.bind(this);
+
+    if (zoomer_ele)
+      this.set_zoomer_element(zoomer_ele);
+  };
+
+  this._set_up_zoomer_ele = function()
+  {
+    var zoomer_ele_style = window.getComputedStyle(this._zoomer_ele);
+    this._zoomer_ele_ver_borders = parseInt(zoomer_ele_style.borderTopWidth) +
```
danfooo, line 84: Safe to rely on `zoomer_ele_style`?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+    if (event.which != 1)
+      return;
+    document.addEventListener("mousemove", this._zoomer_ele_onmousemove_bound);
+    document.addEventListener("mouseup", this._zoomer_ele_onmouseup_bound);
+    var mouse_x = event.clientX - this._zoomer_ele_left;
+    this._overlay_left = mouse_x;
+    this._overlay_right = this._to_right_x(mouse_x);
+    this._mouse_drag_start_x = mouse_x;
+    event.preventDefault();
+    // TODO: need to remove e.g. mousehweel event handlers here
```
danfooo: TODO

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+
+  //
+  // Zoomer element event handlers
+  //
+
+  this._zoomer_ele_onmousedown = function(event)
+  {
+    if (event.which != 1)
+      return;
+    document.addEventListener("mousemove", this._zoomer_ele_onmousemove_bound);
```
danfooo, line 124: useCapture is not optional since very long, or is it?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+    this._overlay_left = mouse_x;
+    this._overlay_right = this._to_right_x(mouse_x);
+    this._mouse_drag_start_x = mouse_x;
+    event.preventDefault();
+    // TODO: need to remove e.g. mousehweel event handlers here
+  };
+
+  this._zoomer_ele_onmousemove = function(event)
+  {
+    document.documentElement.classList.add("overlay-size-change");
```
danfooo, line 135: Does our own `addClass` not work there?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+      var diff = mouse_x - this._to_right_x(this._overlay_right);
+      this._overlay_left = this._mouse_drag_start_x;
+      this.change_overlay_size(0, diff);
+    }
+  };
+
+  this._zoomer_ele_onmousewheel = function(event)
+  {
+    var mouse_x = event.clientX - this._zoomer_ele_left;
+    var diff = (mouse_x < this._overlay_left) ? 5 : -5;
```
danfooo: Should maybe have something CONST for `5`.

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,486 @@
...
+  this._overlay_ele_onmousewheel = function(event)
+  {
+    var diff = (event.wheelDelta < 0) ? 5 : -5;
+    this.change_overlay_size(diff, -diff);
+    event.stopPropagation();
+  };
+
+  this._overlay_ele_onkeydown = function(event)
+  {
+    var width = this._to_right_x(this._overlay_right) - this._overlay_left;
```
danfooo, line 212: I can't really get this to work, also shouldn't it use our standard key handling?

src/network/network_request_crafting_view.js
danfooo, line 200: Quickly checked in 12.02 and 12.10, works fine. But in the end it's always the safest looking things that fail, I'd prefer having it on a bugfix package so we can point to a test that covers it.

src/profiler/profiler.1.0.responses.getevents.js
danfooo: I think it's a bit misleading, one might look at those generated files instead of diving into documentation.

src/profiler/profiler_templates.js
danfooo, line 99: No, true, I was thinking of warnings in the error console. Might add later then, yup.

src/profiler/profiler_templates.js
danfooo: Could there be a toolbar switch to show relative times instead? The zoomed-area scale could then start from 0, so I'd also see how long the events took all together. We'd miss the UI string too, but I think that's less important.

src/profiler/profiler_view.js
danfooo, line 138: Okay, that's what's left for later? Seems okay then.

src/profiler/profiler_view.js
danfooo, line 219: Ah, makes sense.

src/profiler/profiler_view.js
danfooo, line 337: Yes, I think that would help a lot.

src/ui-scripts/zoomer/zoomer.js
hzr: It leads to an annoying behavior where it jumps back and forth on mousewheeling.

src/ui-scripts/zoomer/zoomer.js
hzr, line 84: _zoomer_ele is set before this function is called, so as long as it's set to an element, this is safe. I can add a check that throws if the _zoomer_ele is not set correctly in set_zoomer_element() though.

src/ui-scripts/zoomer/zoomer.js
hzr, line 124: It's been optional for a while.

src/ui-scripts/zoomer/zoomer.js
hzr, line 135: It does, but then it would have a dependency on that. This class is now stand-alone from any other Dragonfly code, so it can be reused without dependencies.

src/ui-scripts/zoomer/zoomer.js
hzr, line 212: The mouse button has to be pressed down. It doesn't use the standard key handling because then it would depend on Dragonfly UI code.

src/ui-scripts/zoomer/zoomer.css
hzr, line 30: I'm not sure what you mean. They are always visible.

src/profiler/profiler_templates.js
hzr: A button was discussed briefly, but I don't like the idea. Some times you want to see one or the other, sometimes you want to see both, but you don't want to press a button to do it.

src/ui-scripts/zoomer/zoomer.css
danfooo, line 30: Its looks like `.zoomer-overlay` is, but the handles are transparent? But what I mean is to style the edge of `.zoomer-overlay` differently on hover (and maybe while dragging).

src/profiler/profiler.1.0.responses.getevents.js
p01: If the 1.1 generated file have these properties, I'd rather go for the generated file then.

src/profiler/profiler_style.css
p01: How bad are double class selectors performance wise ?
If it's really bad, let's switch to `.profiler-event-expandable` and `.profiler-event.non-expandable`

src/profiler/profiler_style.css
```diff
@@ -142,60 +175,63 @@
...
-.profiler-event-interval.selected
+.profiler-event.selected
 {
   box-shadow: 0 0 0 3px #769fe1;
 }
 
 .event-type-3
 {
-  background-color: rgba(199, 199, 235, .6);
+  background-color: rgba(203, 190, 235, 0.6);
```
p01, line 214: Have you notice any impact of using `rgba()` vs `rgb()` or `HEX` colors ?
The timelines can have LOOOOOOOTS of these elements.

src/profiler/profiler_templates.js
```diff
@@ -108,102 +115,76 @@ var ProfilerTemplates = function()
...
+    if (event_list)
     {
       var template = [];
       var total_time = event_list.reduce(function(prev, curr) {
         return prev + curr.time;
       }, 0);
-      event_list.forEach(function(event) {
-        var index = this._order.indexOf(event.type);
+      event_list.forEach(function(event, index) {
         if (index !== -1)
```
p01: This condition is always true, `index` can never be `-1`

src/profiler/profiler_style.css
hzr, line 214: I wouldn't imagine there being a difference between rgb() and hex. rgba() probably means slower painting, but we need it for consistency with other things.

src/profiler/profiler_templates.js
hzr: Oops, leftover code from refactoring, will fix.

src/profiler/profiler_templates.js
```diff
@@ -108,102 +115,76 @@ var ProfilerTemplates = function()
...
-                "class", "profiler-legend-amount"
-               ],
-               ["div",
-                "class", "profiler-legend-amount-bar",
-                "style", "width:" + percentage + "%"
-               ],
-             "class", "profiler-legend-row profiler-timeline-row" + (index % 2 ? " odd" : ""),
+              ["div",
+                ["span",
+                   event_type_string_map[event.type],
```
p01: The extra one space indentation are funky.

hzr: I don't take the "[" into consideration, because it lines up nicer when having attributes, like so:

``` javascript
["span",
   "text",
 "name", "value" 
]
```

So the attribute is on the same level, and the text is indented two spaces from the quotation mark. There are some inconsistencies in this file though. We've never had a standard way of writing templates.

src/profiler/profiler_templates.js
```diff
@@ -211,69 +192,111 @@ var ProfilerTemplates = function()
...
     return template;
   };
 
-  this._timeline_event = function(interval_start, ms_unit, selected_id, event)
+  this._full_timeline_event = function(interval_start, ms_unit, event)
   {
-    var interval = Math.round((event.interval.end - event.interval.start) * ms_unit);
-    var self_time = interval ? Math.max(BAR_MIN_WIDTH, Math.round(event.time * ms_unit)) : 0;
-    var event_start = Math.round((event.interval.start - interval_start) * ms_unit);
+    var width = Math.round((event.interval.end - event.interval.start) * ms_unit);
```
p01: Maybe a small comment about the the template getting a `min-width` from the `.profiler-event-small`declaration.
The possibility of rounding to `0` tingled my spider sense.

src/profiler/profiler_templates.js
```diff
@@ -211,69 +192,111 @@ var ProfilerTemplates = function()
...
+    var left = Math.round((event.interval.start - interval_start) * ms_unit);
```
p01: I didn't see any filtering of the `event_list`, which I suspect from the `interval_start` argument, . Is there any chance `left` can get greater than the container width or that `left`\+ `width` is lower than 0 ?

On second look in the DIFF of the PR, I only see a single call to this private method, with a hardcoded `interval_start` set to 0.

Better drop that argument then.

src/profiler/profiler_templates.js
```diff
@@ -211,69 +192,111 @@ var ProfilerTemplates = function()
...
+  this._timeline_event = function(interval_start, ms_unit, event)
+  {
+    var duration = Math.max(event.interval.end - event.interval.start, MIN_DURATION);
+    var self_time_amount = duration
+                         ? (event.time / duration * 100).toFixed(2)
+                         : 0;
+    var width = Math.round(duration * ms_unit);
+    var left = Math.round((event.interval.start - interval_start) * ms_unit);
     var column = this._order.indexOf(event.type);
     var is_expandable = this._expandables.indexOf(event.type) != -1 && event.childCount > 1;
```
p01: `this._expandables.contains(event.type)`

src/profiler/profiler_templates.js
hzr: Yep, thanks.

src/profiler/profiler_templates.js
hzr: Added a comment.

src/profiler/profiler_style.css
hzr: Yea, I was aware of this one but wanted to profile it. It did show up as one of the top most expensive selectors, so I prefixed it. Thanks!

@p01, they can be quite bad in Opera since they aren't optimized at all.

src/profiler/profiler_view.js
hzr: No, I don't think so, I've changed some stuff.

src/profiler/profiler_style.css
p01, line 214: Makes sense, at worst the color representation would only impact the painting, but that's the lesser problem.

src/profiler/profiler_templates.js
```diff
@@ -211,69 +189,113 @@ var ProfilerTemplates = function()
...
     }
     return template;
   };
 
-  this._timeline_event = function(interval_start, ms_unit, selected_id, event)
+  this._full_timeline_event = function(ms_unit, event)
+  {
+    var duration = event.interval.end - event.interval.start;
+    var width = Math.round(duration * ms_unit); // min-width is specified
+    var left = Math.round((event.interval.start) * ms_unit);
```
p01: Unnecessary parenthesis around `event.interval.start`

src/profiler/profiler_templates.js
```diff
@@ -211,69 +189,113 @@ var ProfilerTemplates = function()
...
+      template.extend(event_list.map(this._timeline_event.bind(this, interval_start, ms_unit)));
     }
     return template;
   };
 
-  this._timeline_event = function(interval_start, ms_unit, selected_id, event)
+  this._full_timeline_event = function(ms_unit, event)
+  {
+    var duration = event.interval.end - event.interval.start;
+    var width = Math.round(duration * ms_unit); // min-width is specified
```
p01: ... is specified _in the style declaration of .profiler-event-small_
or maybe better use the following as this is repeated a couple of times
... is specified _in CSS_

src/profiler/profiler_templates.js
hzr: Thanks, fixed.

src/profiler/profiler_templates.js
hzr: Kinda easy to figure out where, but alright. :)

src/profiler/profiler_view.js
```diff
@@ -349,25 +352,24 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
   this._ontooltip = function(event, target)
   {
-    var id = Number(target.get_attr("parent-node-chain", "data-event-id"));
+    var id = Number(event.target.get_ancestor_attr("data-event-id"));
     var timeline_event = this._timeline_list.get_event_by_id(id);
     if (timeline_event)
     {
-      this._tooltip.show(this._templates.get_title_all(timeline_event));
+      this._tooltip.show(this._templates.get_title_all(timeline_event, this._x0));
     }
```
p01: The curly braces can be omitted here.

src/profiler/profiler_view.js
```diff
@@ -392,43 +407,152 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+    var interval = timeline_list.interval;
+    var visible_event_list = timeline_list.eventList.filter(function(event) {
+      var start = event.interval.start;
+      var end = event.interval.end;
+      return ((start <= x0 && end >= x0) ||
+              (start >= x0 && start <= x1)) &&
+             event.time > this._min_event_time;
+    }, this);
+    var aggregated_event_list = this._get_aggregated_event_list(visible_event_list);
+    var width = this._timeline_width;
```
p01, line 450: That's equivalent to the following and more simple condition ;)

```
return (event.interval.start <= x1 || event.interval.end >= x0)  && event.time > this._min_event_time;
```

hzr, line 450: Thanks! :)

src/profiler/profiler_view.js
```diff
@@ -392,43 +405,149 @@ var ProfilerView = function(id, name, container_class, html, default_handler)
...
+    for (var i = 0, event; event = timeline_list.eventList[i]; i++)
+    {
+      var ele = document.getElementById("profiler-event-" + event.eventID);
+      if (visible_event_list.indexOf(event) != -1)
+      {
+        if (ele)
+        {
+          ele.style.width = Math.round((event.interval.end - event.interval.start) * ms_unit) + "px";
+          ele.style.left = Math.round((event.interval.start - x0) * ms_unit) + "px";
+        }
```
p01, line 464: `visible_event_list.contains(event)`

... or is that another case of being Dragonfly agnostic ?

src/profiler/profiler_view.js
```diff
@@ -496,12 +627,97 @@ cls.Profiler["1.0"].EventList.prototype.get_event_by_id = function(id)
...
+    if (this._table && this._table.get_data())
+      template.extend(this._templates.details(this._table));
+    else
+      template.extend(this._templates.empty(ui_strings.S_PROFILER_NO_DETAILS));
+    this._container.clearAndRender(template);
+  }
+
+  this.ondestroy = function()
+  {
+  };
```
p01, line 685: Is there any advantage in doing:

```
var template = [];
if (foo)
  template.extend(bar);
else
  template.extend(baz);
```

over this:

```
var template;
if (foo)
  template = bar;
else
  template = baz;
```

hzr, line 685: Not really, I just like to provide an initial value.

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,489 @@
...
+      var diff = mouse_x - this._to_right_x(this._overlay_right);
+      this._overlay_left = this._mouse_drag_start_x;
+      this.change_overlay_size(0, diff);
+    }
+  };
+
+  this._zoomer_ele_onmousewheel = function(event)
+  {
+    var mouse_x = event.clientX - this._zoomer_ele_left;
+    var diff = (mouse_x < this._overlay_left) ? MOUSEWHEEL_DIFF : -MOUSEWHEEL_DIFF;
```
p01, line 154:
```
if (event.wheelDelta < 0)
  diff = -diff;
```

hzr, line 154: Yea, I guess that's a bit nicer.

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,489 @@
...
+  };
+
+  /**
+   * Sets the left and right position of the overlay internally. To change
+   * the element position, _update() has to be called.
+   */
+  this._set_overlay_position = function(left, right)
+  {
+    if (left == null)
+      left = 0;
```
p01: Why the `else` ?

OTOH, all the calls to this "private" method have two arguments, so we might skip these tests.

hzr: set_overlay_position(), which is public, calls this directly. Moving the checks there instead (and removing `else`).

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,489 @@
...
+  this.move_overlay = function(diff)
+  {
+    if (!diff)
+      return;
+
+    if (diff < 0)
+      diff = Math.max(this._overlay_left + diff, 0) - this._overlay_left;
+    else
+      diff = Math.min(diff, this._overlay_right);
+    this.change_overlay_size(diff, diff);
```
p01:
```
if (diff != 0)
  this.change_overlay_size(diff, diff);
```

or this can never ever happen ?

src/ui-scripts/zoomer/zoomer.js
```diff
@@ -0,0 +1,489 @@
...
+  };
+
+  /**
+   * Change the size of the overlay.
+   */
+  this.change_overlay_size = function(left_diff, right_diff)
+  {
+    if (left_diff == null)
+      left_diff = 0;
+    else if (right_diff == null)
```
p01: Why the `else` ?

p01:
Looks good beside my last open remarks/comments.

src/ui-scripts/zoomer/zoomer.js
hzr: Removing.

src/ui-scripts/zoomer/zoomer.js
p01: but if `diff == 0` nothing changes, right ? Therefore we don't even need to call this.change_overlay_size`(diff, diff);`

hzr: Ah yea, misunderstood you at first. Thanks.

src/profiler/profiler_view.js
hzr, line 464: No, this is a view class tightly coupled to Dragonfly, so here it's OK. I've changed.

src/ui-scripts/zoomer/zoomer.js
hzr: I don't think this is needed, all event listeners are cleaned up. Removed the comment.

PR #125, opened 2012-11-07
Fix for DFL-3706, Exception in window.templates.resource_tree._get_short_distinguisher

p01:


danfooo:
Have you checked how a resource can not have a url? We should normally instantiate a new resource on urlload, that should have the URL already.

p01:
That's right. I have not seen instances of ResourceInfo without a URL.
OTOH the ListDocuments command of the document-manager has returned documents without a URL.

danfooo:
Is it not the resources that get mocked in resource_service.js:479?

p01:
The `url` and `callback` of cls.ResourceRequest are not optional.
cls.ResourceRequest requests the resource for a given url and calls the callback when the payload comes back.

Also, checking the callstack in DFL-3706, shows it threw when called from window.templates.resource_tree.document(...)

danfooo:
Okay, I see. Looks fine then.

Issue #119, opened 2012-10-31
Add Device-Stock-UA to UA Presets


paddy2k:
The user-agent presets in the "Network options" tab of the Network pane in Dragonfly do not include the Device-Stock-UA header from the latest Opera Mobile 12.1 release. 
https://github.com/operasoftware/Device-Stock-UA-RFC/blob/master/draft-brown-http-device-stock-ua.txt

Similarly, the user-agent presets for Opera Mini should be updated accordingly.

PR #138, opened 2012-11-21
Updated Useragents

paddy2k:
https://github.com/operasoftware/dragonfly/issues/119 - Updated useragents with Device-Stock-UA as per 
https://github.com/operasoftware/Device-Stock-UA-RFC/


p01:
Thanks!

Could you make that pull request into the **bugfixes** branch as we don't commit straight to **master** but to other branches for review before merging into the released branches for testing and incremental releases.

p01:
You might also want to update the AUTHORS file ;)

Issue #153, opened 2013-01-16
Network details doesn't display custom content-type body


mrak:
Our web app uses json with `content-type: application/vendor.corporation.com+json` to communicate between client/server.

This custom content-type is merely plain-text, properly formatted json. The Network details, however, just says:

```
Unable to show data of type application/vendor.corporation.com+json
```

Please show plain-text content/post bodies when possible, or allow a "Show me anyway" option.

This may be related to issue #152

PR #154, opened 2013-01-18
Display custom content-type content bodies in Network Display

mrak:
See issue [#153 Network details doesn't display custom content-type body](https://github.com/operasoftware/dragonfly/issues/153)

I also noticed that some content bodies were displayed starting as `data:application/json;base64,` and then followed by the encoded content. I added in the potential base64 decoding prior to the display in the view as well.

I wasn't sure what the policy was on unit test suite maintenance, so I have not submitted my unit tests in this pull request.

Please let me know if this feature is already completed, in-progress, or acceptable in this form :)


mrak:
No longer in development

PR #170, opened 2014-05-11
SOME FEATURES WE WANT

chiragdkothari:


p01:
Why don't you use the experimental version of Dragonfly instead of forcing all the Dragonfly users to use this version ?

You can set the [URL of (Presto) Opera's developer tool](opera:config#DeveloperTools|DeveloperToolsURL) to **https://dragonfly.opera.com/app/experimental/**

